Issue #7, opened 2026-08-18
Mine negative-space associations: what usually accompanies this change?


teamleaderleo:
## Idea

Most analyzers inspect what exists. Cultist should also notice when a changed construct usually appears together with another construct, but the companion is absent.

This is not a hard rule. It is an empirical association worth asking about.

## Candidate examples

- resource registration usually paired with cleanup / Drop logic
- allocation usually paired with rollback on error paths
- unsafe implementation usually paired with a safety justification
- new parser/decoder usually paired with tests
- new feature flag usually paired with CI coverage
- new dependency capability usually paired with configuration or documentation

Example output:

```text
OBSERVED
29 of 31 analogous registrations also add cleanup behavior.

COUNTEREXAMPLES
2 sites omit cleanup: ...

QUESTION
Does this new registration belong with the exceptions?
```

## Explore

Start with exact syntactic/co-occurrence relationships before attempting semantic data flow. Mine associations at local scopes and require meaningful support before surfacing them.

## Success criteria

- Findings are based on visible support counts and examples.
- Counterexamples are shown.
- The tool does not turn correlation into a claim of correctness.

Issue #4, opened 2026-08-18
Add temporal precedent and convention drift analysis


teamleaderleo:
## Idea

Repository-wide counts can hide migrations. A convention that dominates the full history may be declining, while a newer convention is what maintainers currently prefer.

Cultist should be able to distinguish **historical precedent** from **recent precedent**.

## Explore

For analyzable constructs, attach introduction time / commit information and compare distributions over time:

- all surviving code
- code introduced in the last N months / years
- code introduced since a configurable Git revision or release
- recent changes in the same crate or subsystem

Possible findings:

```text
HISTORICAL PRECEDENT
  unit_tests: 89
  tests:      33

RECENT PRECEDENT (last 24 months)
  unit_tests: 4
  tests:      25

OBSERVATION
  Repository-wide convention and recent convention disagree.
```

## Deterministic evidence

- Git introduction commit/date for occurrences
- counts grouped into time windows
- examples from old and recent cohorts

## Success criteria

- The tool can detect an apparent convention migration without an LLM.
- Time windows and source commits are visible in output.
- Old code does not silently overpower strong recent precedent.

Issue #34, opened 2026-08-18
Build a raw Git co-change explorer


teamleaderleo:
## Idea

Before turning historical coupling into review findings, build research instrumentation that answers a narrower deterministic question:

> Given path A, which paths repeatedly changed with it in comparable history?

This is the first major relational-precedent experiment proposed in PR #33.

## Candidate command

```text
cargo cultist history path/to/file.rs
```

Example output:

```text
HISTORICAL COMPANIONS
  generated/foo.rs    31 / 34 comparable commits
  tests/foo.rs        28 / 34
  Cargo.toml           4 / 34

EXCLUDED
  3 merge commits
  2 repository-wide formatting commits

COUNTEREXAMPLES FOR generated/foo.rs
  abc123  revert
  def456  generator migration
  789abc  ordinary change
```

## First implementation slice

Keep it read-only and exploratory:

- enumerate commits touching a target path;
- collect co-changed paths;
- expose raw support/opportunity counts;
- show representative commits and counterexamples;
- filter obvious merges/reverts/mechanical noise conservatively;
- allow scope limits such as directory/crate when useful;
- preserve the exact commits used/excluded so results are auditable.

Do **not** define review-finding thresholds yet.

## Research targets

Cloud Hypervisor can remain a small sanity target; PR #33 also suggests Codex and Oxc as richer generated/co-change corpora.

## Success criteria

- The command deterministically surfaces plausible historical companions on at least two real repositories.
- Raw and filtered counts are visible.
- Counterexamples are preserved.
- A weak/noisy relationship is allowed to remain weak instead of being forced into a finding.

teamleaderleo:
Completed by merged PR #39. `cargo cultist history FILE` now exposes raw historical companions with support/opportunity counts, representative commits, absence counterexamples, revert/broad-commit filtering, text+JSON output, and explicit limitations. Replay evidence also crossed the proof-of-thesis bar: Oxc `rules.rs` -> generated registry companions appeared in 99/100 considered commits, while the reverse direction was 94/100, showing why the relation model needs direction rather than an undirected co-change edge. Codex replay produced useful noisier companion families and reinforced that cohort selection is the next problem.

Issue #32, opened 2026-08-18
Mine Git history for recurring change companions


teamleaderleo:
## Idea

Some project conventions appear most clearly in history:

```text
when X changes, Y usually changes too
```

The relationship may span files, modules, generated artifacts, tests, configuration, documentation, serializers, protocol tables, or cleanup code. A current diff that changes X without its usual companion Y can therefore be worth a question even when the source tree itself contains no explicit rule.

Cultist should explore **historical change coupling** as repository evidence.

## Examples

Possible learned associations:

- parser changes usually update formatter/serializer cases;
- new enum variants usually update a string table and tests;
- protocol definitions usually update generated bindings;
- feature additions usually update CI coverage;
- dependency feature changes usually update platform configuration;
- public API changes usually update examples or changelog fragments;
- ownership/lifecycle helper changes usually update both success and failure regression tests.

The association is evidence, not an automatic requirement.

## Deterministic first pass

Use local Git history only.

For a bounded recent window:

1. collect changed paths per commit;
2. optionally map changes to Rust items/functions where reliable;
3. count pairwise co-change support;
4. distinguish broad churn commits from focused changes;
5. compare the current diff with strong historical companions;
6. show supporting commits and counterexamples.

Example output:

```text
HISTORICAL PRECEDENT
  Changes touching `FooKind` also changed `foo_to_string()` in 14 of 15 focused commits.

CURRENT DIFF
  `FooKind` changed; `foo_to_string()` did not.

COUNTEREXAMPLE
  One prior commit changed only comments around `FooKind`.

QUESTION
  Does this semantic change require the usual companion update?
```

## Noise controls

Raw file co-change can be garbage in large refactors. Explore filters such as:

- ignore merge/vendor/generated mass updates;
- prefer focused commits below a size threshold;
- weight item-level coupling above file-level coupling;
- separate semantic/code changes from formatting-only commits where detectable;
- require repeated support plus inspectable examples;
- weaken findings when the current diff resembles known counterexamples.

## Relationship to existing issues

- #4 studies how a convention changes over time.
- #7 studies companions visible in the current repository.
- #18 treats Git history as project memory.
- This issue studies **repeated co-change itself as precedent**.

## Dogfood / research seeds

[...]

PR #39, opened 2026-08-18
experiment: mine historical companion paths

teamleaderleo:
## Why

This is the first implementation slice for relational / negative-space precedent.

Instead of immediately declaring a missing companion to be a finding, the experiment exposes the raw historical evidence: given one current file path, which other paths repeatedly changed in the same recent commits?

This directly explores #7 and provides Git/history groundwork for #4.

## What it adds

- `cargo cultist history FILE`
- configurable `--max-commits N`
- text and JSON output
- support counts and percentages
- representative commit examples
- absence counterexamples for top companions
- first-pass filtering for reverts and commits touching more than 100 paths
- explicit limitations around rename history, semantic cohort selection, and correlation
- CI dogfooding against Cargo Cultist's own `src/main.rs`

## Intentional research boundary

This command reports associations before any finding threshold. The next question is empirical: when run against repositories where we already know meaningful source/generated, implementation/test, or other companion relationships, do the useful relationships rise above the noise with credible cohorts?

## First target corpus

The strongest first replay targets from Fieldwork are Codex and Oxc for source/generated and registration/test coupling. WGPU is the next distinct relationship species for lock ordering.


teamleaderleo (commented):
Two things look worth tightening before this experiment becomes the #34/history foundation:

1. The report preserves every excluded commit, but it does not preserve the full set of **considered** commit identities. `considered_commits` is only a count; each companion keeps at most three supporting examples and three absence examples. That means a 31/34 association cannot be independently audited from the JSON packet even though the cohort is the denominator that gives the result meaning. #34 explicitly calls for the exact commits used/excluded to remain auditable. I would carry the complete considered `CommitSummary` list (or an equivalently reproducible cohort receipt) in `HistoryReport`; text can still print only samples.

2. `history-corpus-research.yml` currently runs the full-history Oxc + Codex replay on every pull request targeting `main`. This turns a deliberately external research corpus into routine PR work and adds two large network/history dependencies to every change. The evaluation issue (#16) explicitly leans toward keeping larger external corpus runs separate/on-demand. A `workflow_dispatch` plus the experiment-branch push trigger would preserve the research receipt without making normal PRs pay for it.

The core raw co-change implementation is a useful first probe; keeping the cohort itself as first-class evidence will make later filtering/counterexample work much easier.

teamleaderleo:
Review pass: this is the stronger first explorer and it should remain raw evidence until corpus replay justifies finding thresholds. Two things to preserve before promotion:

1. `HistoryReport` currently imports the shared `REPORT_SCHEMA_VERSION`, but its top-level JSON contract differs from `AnalysisReport` (`anchor`, cohort counts, exclusions, companions, limitations vs `claims`/`findings`). I would decouple the history-experiment schema/version from the stable finding-report schema so version `1` never ambiguously names two contracts. If selected associations later graduate into review findings, emit those through `AnalysisReport` then.
2. Keep the exact pinned Oxc/Codex replay receipts as the next acceptance gate. The external workflow is exactly the right test: inspect strong companions, weak/noisy companions, and counterexamples before deciding any threshold.

The path-resolution behavior in the CLI looks good for the external workflow: the supplied corpus file is canonicalized first, then its own Git root is resolved and the repository-relative anchor is derived. #34 is the raw explorer carrier; #32 can stay the later finding/promotion lane.

teamleaderleo (commented):
Reviewed the exact head `3b7fdcabfd243c31d360cf6a41d40826ab1da3ff` and the external corpus receipts.

The experiment earns its keep. Oxc is the strongest positive control: `crates/oxc_linter/src/rules.rs` co-changes with both generated Rust registry files in 99/100 considered commits, and the lone absence is a docs/license-only change. Codex produces a deliberately messier distribution (README ~80%, combined schemas ~69%, individual protocol surfaces lower), which is useful evidence that the command preserves weak/noisy relationships instead of manufacturing one hidden coupling score.

The code keeps the research boundary clear: deterministic Git facts, visible cohort filtering, examples/counterexamples, schema-versioned JSON, and explicit correlation limitations. CI and the corpus replay are green on this head.

One future hardening item: Git filename transport is line-oriented today; a later cleanup should use NUL-delimited path output so pathological filenames cannot confuse the parser. That does not block this research slice.

I’d land this as the #34 foundation and make the next slice add semantic evidence to the strongest companions rather than adding a finding threshold yet.

teamleaderleo:
Corpus replay result is now durable in `research/history-companion-replay.md`.

Exact extended replay:

- Cargo Cultist: `387d81d9f641b645247e49cfaae5ef327baccae8`
- Oxc: `8783524015b1e6ff1c39ccf426df0bb07cbbc588`
- Codex: `785ecd7452f87c7eb731fbb73892185cbdd9d5f9`
- workflow `32174897832`, job `95834419834`
- artifact `9338659185`
- digest `sha256:69c2afeccb0de9a084204810c1ccf36847a982e47c99d4fa41b1a0a46f6c6115`

Key result: Oxc `crates/oxc_linter/src/rules.rs` co-changed with each core generated registry in **99/100** recent comparable commits. The sole absence was a docs/license-notice change. In the reverse direction, generated `rules_enum.rs` co-changed with source `rules.rs` in **94/100** commits; the missing cases include generated-side performance work.

That asymmetry is useful evidence that the relation model needs direction: `P(B|A)` can represent a strong changed-side expectation even when `P(A|B)` is weaker. Generic symmetric co-change should remain a separate relation kind.

Codex produced the expected second result: real signal with several companion families and a broader cohort. Shared `common.rs` moved with aggregate JSON schemas in ~68-69% of considered commits and the app-server README in ~80%. Narrower `v2/thread.rs` moved with its thread processor in ~80%, README in ~76%, aggregate schemas in ~67%, protocol tests in ~49%, and other consumers at lower rates. This argues for semantic/subsystem cohorts and typed companion relations instead of one global threshold.

Temporary external-corpus workflow has been retired after receipt capture. The experiment remains read-only and keeps raw association evidence separate from future finding thresholds.

Issue #3, opened 2026-08-18
Model precedent by scope and surface precedent tension


teamleaderleo:
## Idea

Treat precedent as scoped evidence instead of one repository-wide popularity score.

A change can simultaneously agree with one scope and disagree with another. The Cloud Hypervisor dogfood already demonstrated this: `unit_tests` was dominant repository-wide while the same `vfio.rs` file already used `tests`.

## Explore

Represent precedent at several scopes where possible:

- same function / impl
- same file
- same module
- same crate/package
- workspace/repository

When those scopes disagree, report **precedent tension** instead of choosing a winner.

Example output:

```text
PRECEDENT TENSION

Repository-wide evidence favors: unit_tests
File-local evidence favors: tests

QUESTION
Which scope is authoritative for this change?
```

## Deterministic evidence

- occurrence counts by scope
- exact examples from each scope
- changed declaration / construct
- whether the local scope already contains another convention

## Success criteria

- The analyzer can explain which scopes agree/disagree.
- It never collapses conflicting evidence into a single hidden score.
- The Cloud Hypervisor `tests` / `unit_tests` case is represented accurately without declaring either name universally correct.

teamleaderleo:
Completed by PR #42 for the first real precedent family. The analyzer now keeps repository-wide and existing file-local evidence separate, emits explicit `PRECEDENT TENSION` when those clear scopes disagree, and reproduced the Cloud Hypervisor #8734 case without declaring either convention universally correct. The replay also excludes the changed declaration from its own precedent baseline (88 existing `unit_tests` vs 33 `tests`). Deeper scope species (crate/subsystem/time/history) are tracked separately in #4, #34, #35, #37, and related roadmap issues.

PR #42, opened 2026-08-18
Surface precedent tension across repository and file scopes

teamleaderleo:
Implements the first slice of #3 for test-module precedent.

The diff analyzer now treats repository-wide and file-local precedent as separate evidence scopes. When both scopes have clear but conflicting conventions, Cultist emits an explicit `PRECEDENT TENSION` instead of making the reader infer the disagreement from counts.

This also tightens the comparison baseline: the changed declaration is excluded from repository precedent counts, so a new pattern cannot vote for itself while being judged.

### Dogfood

A temporary replay against Cloud Hypervisor PR #8734 produced the intended result:

```text
`unit_tests` appears 88 time(s) across 127 existing test-gated modules,
excluding this changed declaration.
Repository precedent counts: `unit_tests`=88, `tests`=33, ...
The same file already uses: `tests`.

PRECEDENT TENSION
  Repository-wide precedent favors: `unit_tests`
  Existing file-local precedent favors: `tests`

OBSERVATION
  Repository-wide and file-local precedent disagree. The change follows
  repository-wide precedent and differs from file-local precedent.
```

This is the exact ambiguity that originally motivated scope-aware precedent. The changed declaration no longer contributes to its own precedent baseline.

The Cloud Hypervisor-only CI step was removed after the replay; permanent CI remains self-dogfood only.

### Validation

- `cargo fmt --check`
- `cargo clippy --all-targets -- -D warnings`
- `cargo test` (11 tests)
- text repository self-scan
- JSON repository self-scan
- text PR diff self-scan
- JSON PR diff self-scan

Examples covered by tests:

- repository-wide `unit_tests`, file-local `tests`, changed `unit_tests` -> tension; change follows repository scope
- tied repository-wide precedent or mixed file-local precedent -> no manufactured tension
- changed occurrence is excluded from precedent counts


teamleaderleo (commented):
There is one behavior hole in the new tension path. `tension` is computed, but the existing early skip still runs on only `different_local_names.is_empty() && !one_off`. For the advertised case where repository precedent favors `unit_tests`, existing file-local precedent is consistently `tests`, and the changed declaration is also `tests`, `different_local_names` is empty and `one_off` is false, so both `build_diff_analysis_report` and `print_diff_report` `continue` before the `PRECEDENT TENSION` branch. That suppresses exactly the “change follows file scope” case described in the PR body.

Please make the skip conditional on there being no tension (for example, include `tension.is_none()`), in both paths, and add an end-to-end/report-level regression where the changed occurrence matches the sole file-local name while repository-wide precedent favors the other name. The helper-only `precedent_tension` test currently passes even though the caller discards that result.

Issue #36, opened 2026-08-18
Cross-check CI test filters against declared tests


teamleaderleo:
## Idea

A CI command can succeed while selecting zero intended tests. Cultist can sometimes prove that deterministically by connecting workflow commands with the repository's declared test inventory.

## First narrow target

Recognize focused Rust test commands such as:

```text
cargo test --lib test_rollback
```

and compare the filter against discovered test names in the relevant package/target where the mapping is unambiguous.

## Candidate finding

```text
PROVEN
  Workflow step runs `cargo test --lib test_rollback`.
  No declared library test contains/matches `test_rollback` under Cargo's filter semantics.

QUESTION
  Is this CI step still exercising the intended regression?
```

## Expand later

- package/target flags
- exact vs substring filtering semantics
- ignored tests
- feature-gated tests
- nextest where deterministic interpretation is practical

## Boundaries

Do not claim zero execution when macros/dynamic harnesses prevent a reliable inventory. Emit `UNKNOWN` or stay quiet.

## Success criteria

- Reproduce at least one known zero-selection-green case from a real repository.
- Keep the first implementation narrowly deterministic.
- Show the workflow location and the test inventory evidence.

teamleaderleo:
Implementation checkpoint from draft PR #46.

Current exact head:

```text
a141c9b69d1357b9a2c55ca5a6b87492b64e6352
```

Authoritative CI run `32177544409` is green. The gate includes:

- rustfmt;
- Clippy with `-D warnings`;
- full unit tests;
- repository text/JSON dogfood;
- history text/JSON dogfood;
- `ci-tests` text/JSON dogfood;
- disposable stale-filter positive fixture;
- matching-filter negative fixture;
- PR diff text/JSON dogfood.

The disposable fixture now verifies the full claim packet:

```text
cargo test --lib stale_filter
-> exactly one ci-test-filter-inventory-miss
-> claim kinds: proven, observed, unknown

cargo test --lib existing_test
-> zero findings
```

The fixture already caught one real analyzer defect during development: ordinary Actions `- run:` list items were initially invisible to the line recognizer. That parser path now has a unit control.

A second review tightened false-positive suppression: only declared Rust module identifiers are accepted as possible libtest qualifier hints. Arbitrary filesystem components such as `src` are no longer treated as evidence that a filter may match a test.

[...]

PR #46, opened 2026-08-18
Experiment with CI test-filter inventory drift

teamleaderleo:
Implements the first deterministic slice of #36:

```text
cargo cultist ci-tests [--format text|json] [PATH]
```

## Evidence model

The analyzer recognizes a deliberately narrow GitHub Actions command family:

```text
cargo [ +TOOLCHAIN ] test --lib FILTER
```

with a conservative set of Cargo flags. It compares each literal selector with:

- explicit `#[test] fn` declarations;
- declared Rust module identifiers, used only as conservative libtest qualifier hints.

A supported filter with no syntax match produces one finding with:

```text
PROVEN   workflow command/filter exists
OBSERVED explicit test/module inventory has no matching name
UNKNOWN  macros/build-time generation may add tests outside syntax inventory
```

Parse failures suppress miss findings. Ambiguous package/integration/bin targets, complex shell forms, qualified filter syntax, unknown flags, and other unsupported forms are skipped instead of guessed.

Text and JSON both render from the shared `AnalysisReport` model.

## Self-dogfood repairs

The permanent disposable CI fixture caught two analyzer mistakes before promotion:

1. ordinary Actions list syntax, `- run: cargo test ...`, was initially missed;
2. arbitrary filesystem path components were initially treated as qualifier hints, which was too permissive.

The final parser handles the list-item form and uses declared module names only.

The permanent fixture requires:

- stale filter -> exactly one `ci-test-filter-inventory-miss` with `proven -> observed -> unknown`;
- matching `existing_test` filter -> zero findings.

## External control: exrs

`johannesvollmer/exrs@63b4ef472537d38374d9856f469f16d5b1fcc714` explicitly asserts an expected passing-test count for filtered SIMD runs because a zero-match libtest selector can still exit successfully. That matrix/direct-binary form remains outside this literal Cargo parser, but it independently validates the failure class.

## Real acceptance witness: Tantivy

Exact retained carrier:

```text
teamleaderleo/tantivy@b92909ef3d5ac5695d1c85b1b0cb52a03ee51e49
.github/workflows/fieldwork-prepare-commit-generation-fence.yml
```

Literal selector:

```text
cargo +1.88.0 test --lib test_rollback --locked --no-default-features
```

Historical execution:

```text
workflow run: 30513367302
job:          90777910777
conclusion:   success
```

The command reported:

```text
running 0 tests
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 1120 filtered out
```

while the job stayed green. The adjacent `test_prepare_` control selected and passed three tests.

Cargo Cultist replayed the exact pinned carrier at head `3a32359a19be24819f54e2ab0648d52d88ddb19f` and required exactly one `test_rollback` finding at that workflow path. Replay run `32216675650` passed; generic CI `32216675619` passed.

The temporary Tantivy network workflow has since been removed. The full receipt is retained in `research/ci-test-filter-replay.md`.

## Final validation

Exact cleanup head:

```text
19618f42adb6f6833dcd3bfe8c916fdc924ac7bb
```

Generic CI run `32216810487` passed after the temporary replay workflow was retired and CLI help was aligned with the final evidence model.

## Result

The first-slice success criteria from #36 are met:

- real green zero-selection case reproduced and pinned;
- first implementation remains narrowly deterministic;
- workflow location and syntax-inventory evidence are present;
- runtime/generated uncertainty remains explicit.

Execution-aware verification is continuing separately in stacked PR #58 so ordinary `ci-tests` remains read-only.

Refs #36.


teamleaderleo (commented):
Two things before this experiment is ready to evaluate against an external zero-selection case.

1. **Cargo filters match the full test name, not only the function identifier.** The current inventory stores `ItemFn.sig.ident` and checks `test.name.contains(filter)`. That creates a clean false question for module-level filters. Example:

```rust
mod recovery {
    #[test]
    fn rollback() {}
}
```

`cargo test --lib recovery` selects `recovery::rollback`, while this analyzer reports zero explicit name matches because `rollback` does not contain `recovery`.

I’d make the explicit inventory retain the recoverable harness-style path (at minimum inline module stack + function name) and compare the filter against that full name. Where module identity cannot be reconstructed confidently, keep it UNKNOWN / outside the supported slice. Add a reversal test for the example above.

2. Current CI run `32175407932` fails only at `cargo fmt --check`; the formatter diff is mechanical in `src/ci_test_filters.rs`.

The claim taxonomy is otherwise well judged: exact recognized command as PROVEN, syntax inventory miss as OBSERVED, macro/build-time coverage as UNKNOWN. Keeping parse failures suppressive is the right posture for this first slice.

teamleaderleo:
External control found: `johannesvollmer/exrs@63b4ef472537d38374d9856f469f16d5b1fcc714` explicitly defends against this class in `.github/workflows/rust.yml`.

Its SDE job builds the lib test binary, runs it directly with a matrix name filter, and then checks the exact expected passing-test count. The workflow comment states the reason directly: a name filter that matches nothing still exits successfully with `0 passed`, so a typo/stale filter could otherwise produce a green step.

This is a useful **negative/control case**, not a case #46 should forcibly parse. The filter is passed to the compiled test executable through matrix variables, outside the deliberately narrow `cargo test --lib FILTER` command family implemented here. That boundary is valuable: the broader relation is `CI selector -> resolved test inventory / executed count`, while this first syntax slice owns only literal Cargo-filter commands.

Research consequence:
- keep this PR narrow and precision-biased;
- use exrs as evidence that expected-count assertions are one legitimate repository-local defense;
- recover a real literal-Cargo zero-selection case before calling #36 accepted;
- a later execution-aware analyzer can recognize direct test-binary filters / wrappers and use actual listing or result-count evidence instead of guessing through shell dataflow.

teamleaderleo:
Follow-up review: the module-filter concern is addressed conservatively in current head `a141c9b69d1357b9a2c55ca5a6b87492b64e6352`.

The analyzer now records declared `ItemMod` names as possible libtest qualifier fragments, so a filter such as `recovery` is suppressed when the source declares `mod recovery`, while arbitrary source-path fragments such as `src` do not count. That trades some recall for precision, which is the right direction for this experiment. The targeted unit control covers the distinction, and exact-head CI run `32177544409` is green.

I agree with keeping #46 draft despite the clean implementation: the remaining gate is empirical, not code hygiene. The exrs example proves the broader zero-selection failure class exists and also demonstrates one repository-local defense, but it intentionally sits outside this parser family. Promote this slice only after recovering a real literal `cargo test --lib FILTER` stale-selector case or after an execution-aware successor supplies authoritative selection evidence.

Issue #51, opened 2026-08-18
Stream Git output and represent changed lines as compact ranges


teamleaderleo:
## Why

Several current paths materialize work at the most expensive granularity:

- `git_diff` captures the entire patch into a `Vec<u8>`/`String` via `Command::output()`;
- `parse_changed_lines` inserts every added line number into a `BTreeSet<usize>`;
- the history experiment captures each complete `git show` output before parsing and only then decides whether a commit was too broad.

A large generated diff or mass-change commit can therefore consume memory proportional to patch bytes plus one tree node per added line, even though Cultist mostly needs changed paths and hunk membership.

## Direction

Parse Git output as a stream and retain compact evidence.

### Diff

- spawn Git with piped stdout and consume through a buffered reader;
- restrict pathspecs to the active analyzers' relevant file classes (#43);
- represent added lines as merged ranges/hunks per path instead of individual `usize` entries;
- membership becomes interval search, with storage proportional to hunks rather than added lines;
- keep exact path/line semantics for findings.

Example internal representation:

```text
src/a.rs -> [11..=12, 32..=32]
```

rather than `{11, 12, 32}` in a tree set.

### History

When reading changed paths for a commit/batch:

- count paths while streaming;
- once a broad-commit threshold is crossed, stop retaining individual paths for that commit while continuing enough parsing to preserve exact counts/record boundaries;
- avoid holding raw Git output after it has been converted into facts.

### Rendering

If machine-readable reports become large, prefer `serde_json::to_writer[_pretty]` to a locked/buffered stdout over building one giant JSON `String` first. Apply the same principle to large text output where profiles justify it.

## Success criteria

- diff memory scales with changed paths/hunks rather than patch byte size or number of added lines;
- a synthetic/fixture million-line contiguous addition is represented by O(1) hunk/range entries for that run;
- history broad commits can be excluded without retaining every path string;
- existing diff membership tests and merge-base semantics remain identical;
- peak RSS benchmark is added for a large generated diff and broad history commit;
- JSON output can be emitted without an output-sized intermediate `String`.

## Related

- #43 changed-first diff execution
- #47 batched history extraction
- #48 performance corpus
- #49 performance program

teamleaderleo:
Current-main handoff after merged #351 and #353:

Delivered in #51's owned lane:
- history Git output streams incrementally;
- broad commits stop retaining path strings after the threshold while preserving exact changed-path counts;
- diff Git output streams through bounded readers;
[...]

Issue #49, opened 2026-08-18
Performance program: demand-driven work, incremental reuse, one-pass evidence


teamleaderleo:
## Goal

Make Cultist cheap enough to live in the edit/review loop even as repository reasoning gets much richer.

The guiding target is **work proportional to the evidence actually needed**:

- no-op / irrelevant diffs should approach startup + Git-diff cost;
- warm analysis should approach the size of the changed content, not the repository;
- cold full analysis should approach one inventory/read/extraction pass;
- multiple analyzers should share facts instead of multiplying scans;
- history analyzers should share batched/indexed Git facts instead of multiplying Git processes.

## Current evidence

Main currently walks every Rust file, reads it, and runs a full `syn` parse for the test-module analyzer before diff relevance is known.

The new CI-test-filter experiment in PR #46 independently adds another whole-repository `WalkDir` + `read_to_string` + `syn::parse_file` pass to discover explicit tests. That is a useful warning: analyzer growth is already beginning to multiply repository work.

The history experiment in PR #39 currently performs one Git process to discover commits and then one `git show` process per commit.

These are good prototype choices. The performance lane is about changing the execution model before they become permanent scaling behavior.

### 1. Measure the product loop

#48 — end-to-end performance corpus, internal stage counters, and regression budgets.

Measure cold/warm/no-op/diff/history scenarios plus CPU, RSS, source files opened/parsed, cache hits, and Git subprocess count.

### 2. Make diff/review execution demand-driven

#43 — inspect changed paths/evidence first and short-circuit analyzers that cannot produce a finding.

This principle should generalize. Example for PR #46: parse supported workflow commands first; if none exist, skip the Rust test inventory entirely.

### 3. Reuse facts by content identity

#44 + #13 — use Git blob identity for clean tracked files, hash only dirty/untracked content, and cache facts by content + extractor version.

A warm run with cached facts should be able to avoid opening unchanged source files.

### 4. Extract once, at the cheapest sufficient tier

#45 — conservative byte gates -> lightweight syntax facts -> full parse -> semantic adapter only when requested.

One file/content parse should feed every active analyzer that needs those facts. Parallelize CPU-heavy extraction after filtering/cache lookup; sort at the output boundary for determinism.

### 5. Make history a shared evidence source

#47 — batch Git history extraction immediately, then build reusable commit<->path facts without quadratic precomputation.

## Tactical optimizations to measure after the execution model lands

Keep these as experiments until profiles say they matter:

[...]

Issue #48, opened 2026-08-18
Add end-to-end performance corpus, stage timings, and regression budgets


teamleaderleo:
## Why

Performance is becoming a product property for Cultist. The tool will sit in local edit/review loops, CI, and repeated archaeology queries; a correct analyzer that makes every question pay repository-scale work will feel wrong.

Current CI dogfoods text and JSON repository scans plus diff modes, but it does not measure execution cost. #16 already includes runtime/indexing cost in evaluation; make that a dedicated repeatable performance corpus.

## Benchmark scenarios

Measure the cases developers will actually feel:

- process startup / help;
- full cold repository scan;
- full warm repository scan with an evidence cache;
- no-op diff;
- docs-only diff;
- small irrelevant Rust diff;
- small relevant Rust diff;
- one-file incremental rescan;
- branch switch with mostly shared blobs;
- history query at 100 / 1,000+ commits;
- repeated history queries in one repository.

Use small, medium, and large Rust repositories. Reuse stable dogfood/research corpora from #16/#33 where practical rather than inventing synthetic-only benchmarks.

## Measure more than wall time

Capture at least:

- wall-clock latency;
- user/system CPU;
- peak RSS;
- number of source files opened/parsed;
- cache hit/miss counts;
- Git subprocess count;
- bytes or files processed when cheaply observable.

A run can get faster by consuming absurd memory or hiding work in subprocesses; the benchmark should expose that.

## Stage timings

Add low-overhead internal spans/counters behind an opt-in flag/environment variable so slow runs explain themselves. Candidate stages:

- repository inventory;
- Git diff/history collection;
- cache lookup;
- file read/hash;
- byte/lexer/full parse extraction;
- semantic adapter time;
- analyzer/reduction time;
- rendering/serialization.

Keep ordinary output clean and deterministic.

## Regression policy

1. establish baselines first;
2. publish benchmark deltas on performance-sensitive PRs;
3. use generous/noisy-CI-aware thresholds initially;
4. fail only on large stable regressions once variance is understood;
5. retain raw benchmark artifacts for comparison.

Absolute latency targets can be set after baseline data exists. The design goal is that warm/no-op/irrelevant-diff work scales with the changed/relevant set, while cold full analysis scales close to one repository read/extraction pass.

## Success criteria

- one command/workflow reproduces the benchmark matrix;
- performance counters identify where time/work went;
- cold and warm results are distinct;
- PRs changing scan/index/history code can show before/after data;
[...]

Issue #45, opened 2026-08-18
Build a tiered Rust fact extractor instead of full-parsing every file


teamleaderleo:
## Why

The first analyzer currently enables `syn`'s `full` feature and runs `syn::parse_file` for every Rust file even though the requested fact is narrow: test-gated module declarations.

As more analyzers arrive, the expensive failure mode is either:

- every analyzer performs its own source pass; or
- every file is always promoted to the deepest parser/semantic backend because some analyzer might need it.

Both make cheap questions expensive.

## Direction

Introduce explicit extraction tiers and promote files only as far as the requested facts require.

Possible tiers:

1. **byte prefilter** — conservative substring/signature checks used only to rule files *in* for more work;
2. **lexer / lightweight syntax events** — attributes, item boundaries, identifiers, literals/comments where reliable;
3. **full syntax parse** — `syn` or another syntax backend for analyzers that need an AST;
4. **semantic adapter** — rust-analyzer/rustc-backed facts only for the selected files/items that require resolution/types/references.

A false positive in a cheap tier only costs work. A false negative that hides a valid fact is unacceptable, so uncertain cases must promote to the next tier.

## Parse once, feed many analyzers

Extraction should produce a shared compact fact/event stream. Active analyzers consume those facts instead of each visiting the same AST independently or triggering their own repository scan.

For syntax-heavy runs, parse each content blob once per extractor version and derive all requested syntax facts in that pass.

## Parallelism

After relevance filtering/cache lookup, CPU-heavy extraction is a good parallel stage. Use bounded worker parallelism and merge into deterministic output at the end. Determinism should come from explicit sorting/rendering, not from paying `BTreeMap`/`BTreeSet` costs in every hot-path accumulator.

## First experiment

For test-module discovery, compare:

- current full `syn` parse of every Rust file;
- conservative byte prefilter followed by full parse only for candidates;
- a narrow lexer/event extractor if it can preserve correctness cleanly.

Measure candidate rate, parse time, allocation/RSS, and total end-to-end latency on small/medium/large Rust repositories.

## Success criteria

- multiple analyzers can request facts through one extraction API;
- a file is promoted only to the deepest evidence tier actually required;
- a syntax file is parsed at most once per content/extractor version in a run;
- deterministic results survive parallel extraction;
- benchmark data decides whether a dedicated lightweight Rust syntax extractor earns its complexity;
- semantic backends remain optional and targeted.

## Related

- #13 local evidence index
- #14 semantic adapters
[...]

PR #55, opened 2026-08-19
Make diff analysis changed-first

### LINKED ISSUE #43: Make diff analysis changed-first and short-circuit irrelevant work
## Why

`run_diff` currently calls `analyze_test_modules(&root)` before the diff is inspected. `analyze_test_modules` walks the repository, reads every Rust file, and runs a full `syn::parse_file` on each one.

That means a docs-only change, or a Rust change with no test-module declaration, still pays repository-scale parse cost before Cultist can say there is nothing relevant to this analyzer.

For a diff-oriented review tool, the common cheap case should be extremely cheap.

## Direction

Invert the pipeline:

1. resolve the Git root / merge base;
2. inspect the diff first, path-limited to the file kinds required by active analyzers;
3. identify changed paths/hunks;
4. run the cheapest changed-file trigger needed to determine which analyzers can possibly produce findings;
5. parse only changed candidate files;
6. request repository-wide precedent facts only for analyzers that still have a candidate finding.

For the current test-module analyzer, if no changed Rust file contains an added/renamed test-gated module declaration, return immediately without walking/parsing the repository.

Longer term, analyzers should declare their input requirements / trigger conditions so the engine can build a small execution plan instead of every analyzer independently requesting a full repository scan.

## Important semantic constraint

The optimization must preserve current merge-base and working-tree semantics, including staged and unstaged changes.

Cheap triggering should be conservative: false positives may cause extra work; false negatives that skip a real finding are unacceptable.

## Success criteria

- docs-only diff performs no repository Rust parse;
- irrelevant Rust diff avoids a repository-wide parse;
- relevant diff preserves current findings and output;
- Git diff is restricted to relevant path classes when possible;
- add end-to-end benchmarks for no-op/docs-only/irrelevant-Rust/relevant-Rust diffs on at least one large Rust repository;
- report cold and warm timings separately.

## Related

- #13 local evidence index
- #16 evaluation corpus
- #3 scoped precedent

teamleaderleo:
## What

Make the current diff analyzer pay repository-wide parsing cost only after the diff proves it has a relevant changed test-module declaration.

The execution path is now:

1. resolve the diff, path-limited to Rust files;
2. retain added-line membership as compact ranges;
3. parse only changed Rust files;
4. return immediately when none of their changed lines contain a test-gated module declaration;
5. only then scan the rest of the repository for precedent, reusing the changed-file parse results.

The repository-wide command keeps the same full scan behavior.

## Regression coverage

Two end-to-end tests put an invalid-UTF-8 Rust file elsewhere in a temporary Git repository:

- a docs-only diff must succeed without touching repository Rust source;
- an unrelated Rust diff must parse only the changed Rust file and succeed without touching the unrelated source.

That makes the changed-first property observable in tests.

The diff representation also coalesces contiguous added lines into ranges instead of storing one `BTreeSet` node per added line. Streaming the Git patch remains follow-up work in #51.

## Why

This is the first execution-model change from the performance program: cheap and irrelevant review paths should do work proportional to the diff, while repository-wide precedent is fetched only for candidate findings.

Closes #43
Progresses #51
Related: #49 #48 #45

### DIFF
--- a/src/diff.rs
+++ b/src/diff.rs
@@ -4,32 +4,59 @@ use std::path::{Path, PathBuf};
 use std::process::Command;
 
 use crate::finding::{AnalysisReport, Claim, ClaimKind, Evidence, Finding, Location};
-use crate::test_modules::{TestModuleOccurrence, TestModuleReport};
+use crate::test_modules::{
+    TestModuleOccurrence, TestModuleReport, analyze_test_module_files,
+    analyze_test_modules_excluding,
+};
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+struct LineRange {
+    start: usize,
+    end: usize,
+}
 
 #[derive(Debug, Default, Eq, PartialEq)]
 struct ChangedLines {
-    by_path: BTreeMap<PathBuf, BTreeSet<usize>>,
+    by_path: BTreeMap<PathBuf, Vec<LineRange>>,
 }
 
 impl ChangedLines {
     fn insert(&mut self, path: &Path, line: usize) {
-        self.by_path
-            .entry(path.to_path_buf())
-            .or_default()
-            .insert(line);
+        let ranges = self.by_path.entry(path.to_path_buf()).or_default();
+        if let Some(last) = ranges.last_mut()
+            && line <= last.end.saturating_add(1)
+        {
+            last.end = last.end.max(line);
+            return;
+        }
+        ranges.push(LineRange {
+            start: line,
+            end: line,
+        });
     }
 
     fn contains(&self, path: &Path, line: usize) -> bool {
-        self.by_path
-            .get(path)
-            .is_some_and(|lines| lines.contains(&line))
+        self.by_path.get(path).is_some_and(|ranges| {
+            ranges
+                .iter()
+                .any(|range| range.start <= line && line <= range.end)
+        })
     }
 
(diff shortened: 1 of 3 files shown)


teamleaderleo (commented):
Direction is strong: the changed-first execution model is exactly the right performance move, and the invalid-UTF-8 controls make the avoided repository scan observable.

Two things before integration:

1. Current exact-head CI run `32215842050` stops at `cargo fmt --check`; the reported diffs are mechanical in `src/diff.rs`.

2. Preserve uncertainty when a **changed** Rust file parses as UTF-8 but fails `syn::parse_file`. Right now `analyze_test_module_files()` records that parse failure, then `changed_test_modules(...).is_empty()` takes the same early clean path as a successfully parsed irrelevant diff and emits `No added or renamed test-gated module declarations were found in the diff.` In that case the changed-first scan did not establish absence; it failed to inspect the changed syntax.

This was easy to overlook because the old repository-wide path also carried weak parse-failure handling, but the new early-return boundary is a good place to make the evidence contract explicit. I’d add a changed-file syntax-error fixture and either propagate an UNKNOWN claim / parse note and stop, or return an analysis error. The key is: parse failure must not become an OBSERVED absence claim.

Once that and formatting are clean, I’m happy with the changed-first design. Keeping contiguous added lines as ranges is a sensible local memory improvement and doesn’t need a grander abstraction yet.

PR #58, opened 2026-08-19
experiment: verify CI test selectors with libtest listing

teamleaderleo:
Execution-aware follow-up to merged #46 / issue #36.

## Question

Can Cargo Cultist replace the syntax-only `UNKNOWN` with authoritative test-selection evidence for the narrow supported `cargo [ +TOOLCHAIN ] test --lib FILTER` family?

## Safety boundary

This is deliberately **opt-in execution research**. Compiling a Cargo test target can execute repository build scripts. Ordinary `cargo cultist ci-tests` remains read-only and syntax-only.

The probe reconstructs the supported Cargo invocation and adds libtest listing:

```text
cargo [ +TOOLCHAIN ] test ... FILTER -- --list
```

Cargo therefore resolves/builds the same library test target while libtest lists selected tests without executing those selected tests.

## Synthetic discriminator

A disposable Cargo repository contains one explicit test, `tests::existing_test`, while its workflow says:

```text
cargo test --lib stale_filter
```

The research workflow must prove three facts independently:

1. ordinary stale filtered execution exits successfully while reporting `0 passed`;
2. libtest listing reports `listed selections: 0`;
3. changing only the filter to `existing_test` reports exactly one selected test, `tests::existing_test`.

## Real replay target: Tantivy

The same workflow pins the accepted #46 corpus carrier:

```text
teamleaderleo/tantivy@b92909ef3d5ac5695d1c85b1b0cb52a03ee51e49
```

and installs Rust 1.88.0 so the exact workflow selectors can be reconstructed:

```text
test_prepare_  -> expected 3 selected tests
test_rollback  -> expected 0 selected tests
```

This gives the execution-aware probe a real external control paired with the historical run where `test_rollback` stayed green with zero tests.

## Current code

`examples/ci_test_selection_listing.rs` reuses the merged #46 command discovery, preserves a simple Cargo toolchain selector, converts supported commands to `--list`, and parses selected test/benchmark names.

This remains research code until execution behavior, build-script risk, command reconstruction, and Tantivy replay are green on the current main base.


teamleaderleo (commented):
This is the right successor direction, and the Tantivy carrier is a real upgrade over the synthetic-only cell.

The pinned Tantivy commit `b92909ef...` itself contains the literal supported commands and intentionally generates a lockfile before executing them, so reproducing that sequence follows the carrier's declared workflow model. If the replay reaches:

```text
test_prepare_ -> listed selections: 3
test_rollback -> listed selections: 0
```

then #36 finally has a real external literal-`cargo test --lib FILTER` zero-selection case.

### Keep one epistemic distinction explicit

Authoritative libtest listing resolves **selection existence** for the compiled target at this checkout/toolchain/dependency resolution. It does not by itself establish **intent**.

So a future finding packet should separate:

```text
PROVEN / EXECUTED
  Current resolved libtest target lists zero tests for `test_rollback`.

OBSERVED / HISTORICAL / EXPLICIT
  Nearby workflow controls or prior repository evidence indicate what this selector was intended to cover.

UNKNOWN
  Zero selection alone does not prove that an empty selector is erroneous or stale.
```

That preserves the core Cultist rule even when execution removes the syntax-only uncertainty.

### Effect boundary

The warning about build scripts is essential. `cargo test ... -- --list` can compile dependencies, run build scripts/proc macros, write `target/`, update caches, and contact registries. Keep this as explicit opt-in execution evidence; ordinary `ci-tests` should stay read-only.

### Current failure

Research run `32216696618` has not reached the synthetic or Tantivy replay yet. `cargo clippy --example ... -D warnings` fails because the example includes the full production modules with `#[path = ...]`, producing dead-code errors for functions/types unused by the example. Generic CI is red as well.

I’d fix that by extracting the narrow command-discovery pieces into a reusable module or giving the research example a purpose-built parser interface. A blanket dead-code allow over the included production module would make the experiment run, but it would also weaken the example's own quality gate.

Once the real Tantivy replay passes, record its exact dependency/toolchain/run receipt. The generated lockfile means the replay is authoritative for that executed graph, not a byte-identical reconstruction of the historical graph.

Issue #38, opened 2026-08-18
Treat explicit repository guidance as higher-authority precedent evidence


teamleaderleo:
## Idea

Observed precedent is useful, but sometimes the repository has already written the rule down. Cultist should discover explicit guidance and surface when it agrees or conflicts with mined precedent.

## Candidate sources

- `AGENTS.md`
- `CONTRIBUTING.md`
- developer docs
- generator headers/comments
- workflow comments
- lint/config policy
- Cargo metadata
- scripts with explicit generation/test instructions

## Desired behavior

```text
EXPLICIT
  CONTRIBUTING.md says generated registries must be updated through `cargo lintgen`.

OBSERVED
  37/38 comparable rule additions follow that generator path.

CHANGE
  Generated registry edited directly.

QUESTION
  Was the direct edit intentional?
```

When explicit guidance conflicts with historical frequency, show the disagreement instead of silently choosing one.

## Boundaries

The first version can use explicit machine-readable/strongly linked instructions. General natural-language policy extraction can remain optional or model-assisted later.

## Success criteria

- At least one analyzer can attach explicit repo guidance as evidence.
- Explicit and observed precedent remain separately labeled.
- Contradictory guidance/history becomes a visible tension rather than a hidden weighting decision.

teamleaderleo:
Research note from 2026-08-19: this issue should treat explicit guidance as **higher-authority evidence with freshness/provenance**, not timeless truth.

Two current signals reinforce that boundary:

- GitHub Copilot code review now explicitly consumes repository instructions (`AGENTS.md`, `.github/copilot-instructions.md`, path-specific instructions, skills) as review context. That makes guidance files an increasingly real execution input, not just documentation: https://docs.github.com/en/copilot/concepts/agents/code-review
- *The Working Set of a Coding Agent: Coherence Debt in Repository-Scale Tasks* (arXiv:2608.16630, 2026-08-17) reports that when repository guidance and current code disagree, agents can follow the stale guidance even when it produces worse code: https://arxiv.org/abs/2608.16630

Concrete Cultist consequence: preserve at least three separate claims when guidance is involved:

```text
EXPLICIT
  Current guidance says X.

OBSERVED
  Recent comparable changes/code follow Y.

TEMPORAL
  Guidance X was introduced/superseded at commit/date Z.
```

[...]

Issue #35, opened 2026-08-18
Detect missing or directly edited generated companions


teamleaderleo:
## Idea

Use explicit generator evidence plus historical coupling to identify generated artifacts that usually move with source changes.

This should build on #34 instead of treating co-change alone as proof of generation.

## Evidence sources

- generated-file headers such as `@generated` / `DO NOT EDIT`
- generator scripts and repository commands
- comments naming source inputs or generator ownership
- source/output path conventions
- historical co-change
- deterministic regeneration when cheap/safe

### Missing regeneration

```text
FACTS
  src/rules.rs changed.
  generated/registry.rs is explicitly marked generated.
  42/44 comparable source changes updated it.

COUNTEREXAMPLES
  1 revert
  1 generator migration

QUESTION
  Is regeneration intentionally deferred?
```

### Direct edit of generated output

```text
FACTS
  generated/registry.rs changed.
  Declared generator inputs did not.
  Comparable changes normally update this file through generator output.

QUESTION
  Was this generated file intentionally edited directly?
```

## Success criteria

- Require at least one explicit generator/generated signal in addition to history before calling something a generated companion.
- Show source/output evidence and counterexamples.
- Produce one useful real-repository finding with high precision.

teamleaderleo:
#39 has now landed the raw history foundation, and its replay gives #35 a concrete first target.

## First implementation slice

Keep the next step evidentiary instead of jumping straight to a missing-regeneration finding:

```text
historical companion
+ explicit generated-file marker
-> annotated companion evidence
```

Proposed deterministic behavior:

- for current companion paths that still exist as regular files, scan only a bounded header prefix;
- recognize strong explicit markers such as `@generated`, `DO NOT EDIT`, `automatically generated`, `auto-generated`;
- retain the exact marker text and source path as evidence;
- expose that evidence in text + JSON history output;
- keep the relationship as historical association; the marker proves the file declares itself generated, while it still does **not** prove which source/generator owns it;
- add an explicit limitation that generated-marker evidence alone does not establish derivation direction or a regeneration requirement.

The Oxc replay in `research/history-companion-replay.md` is the first real target: `crates/oxc_linter/src/rules.rs` moved with both generated Rust registries in 99/100 considered commits, and the reverse direction was only 94/100. That is a clean case for combining explicit generated evidence with **directional** history rather than one symmetric coupling score.

[...]

PR #59, opened 2026-08-19
experiment: recover Rust generator ownership from literal path I/O

teamleaderleo:
## Question

Can Cargo Cultist recover a real source -> generated-output ownership relation from deterministic repository code, without naming the target repository in the analyzer?

This follows the merged historical-companion, generated-marker, and Rust-syntax-cohort work.

## Experiment

Retains a standalone `examples/rust_generator_ownership.rs` research probe. For one Rust generator source file it:

- parses functions with `syn`;
- recognizes literal repository paths passed through `.join("...")`;
- tracks simple local path bindings;
- records `fs::read` / `fs::read_to_string` inputs and `fs::write` outputs in the same function;
- finds the nearest Cargo package name;
- finds Cargo aliases whose command names that package;
- independently checks exact output paths marked `linguist-generated=true` in `.gitattributes`.

Dynamic paths and broader dataflow remain outside this first adapter.

## Oxc acceptance replay

Pinned target:

```text
oxc-project/oxc@8783524015b1e6ff1c39ccf426df0bb07cbbc588
```

Generator:

```text
tasks/linter_codegen/src/main.rs
```

Executed research head `5192a4aae16f2ae9e8d5f670abd969912cae03fb` passed workflow run `32216773896`, job `95959664054`; artifact `9352598772`, digest `sha256:8147a985fef292a5c4eda48764b4fec86bb58feb738dbfbec643f6a65eab7cb4`.

The probe recovered:

```text
generator package: oxc_linter_codegen
cargo lintgen -> run -p oxc_linter_codegen

function generate_rule_runner_impls
  reads  crates/oxc_linter/src/rules.rs
  writes crates/oxc_linter/src/generated/rule_runner_impls.rs
         [.gitattributes: linguist-generated=true]

function generate_rules_enum_file
  reads  crates/oxc_linter/src/rules.rs
  writes crates/oxc_linter/src/generated/rules_enum.rs
         [.gitattributes: linguist-generated=true]
```

The temporary network workflow was removed after the receipt. Full evidence is preserved in `research/rust-generator-ownership-replay.md`.

## Evidence boundary

A literal read/write pair establishes repository-code data movement and a typed generator-input/output relation. It does **not** say every possible input edit changes emitted bytes. Historical cohort relevance and the changed fact remain separate evidence.

## Final cleanup

The branch was compacted to one two-file commit on top of the current main that already contains merged #58:

```text
172d0480e1da8340786ea2630381be953a7259a3
```

Generic CI run `32217239781` passed.

The next composition experiment is #66: combine current diff absence + explicit guidance + generator ownership + generated markers + the 99/99 syntax cohort into one bounded evidence packet.

Refs #35, #38.

### DIFF
--- a/examples/rust_generator_ownership.rs
+++ b/examples/rust_generator_ownership.rs
@@ -1,0 +1,447 @@
+use std::collections::{BTreeMap, BTreeSet};
+use std::env;
+use std::error::Error;
+use std::fs;
+use std::path::{Path, PathBuf};
+
+use syn::visit::{self, Visit};
+use syn::{Expr, ExprCall, ExprLit, ExprMethodCall, File, ItemFn, Lit, Pat, Stmt};
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+struct FunctionIo {
+    function: String,
+    reads: BTreeSet<String>,
+    writes: BTreeSet<String>,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+struct CargoAlias {
+    name: String,
+    command: String,
+}
+
+fn main() {
+    if let Err(error) = run() {
+        eprintln!("rust-generator-ownership: {error}");
+        std::process::exit(1);
+    }
+}
+
+fn run() -> Result<(), Box<dyn Error>> {
+    let mut args = env::args().skip(1);
+    let root = PathBuf::from(
+        args.next()
+            .ok_or("usage: rust_generator_ownership REPO GENERATOR_RS")?,
+    )
+    .canonicalize()?;
+    let generator = PathBuf::from(
+        args.next()
+            .ok_or("usage: rust_generator_ownership REPO GENERATOR_RS")?,
+    );
+    if args.next().is_some() {
+        return Err("usage: rust_generator_ownership REPO GENERATOR_RS".into());
+    }
+    if generator.is_absolute() {
+        return Err("GENERATOR_RS must be repository-relative".into());
+    }
+
+    let generator_path = root.join(&generator);
+    let source = fs::read_to_string(&generator_path)?;
+    let file = syn::parse_file(&source)?;
+    let io = collect_function_io(&file);
+    let package_name = nearest_package_name(&root, &generator_path);
+    let aliases = package_name
+        .as_deref()
+        .map(|package| cargo_aliases_for_package(&root, package))
+        .unwrap_or_default();
+    let generated_attrs = generated_attribute_paths(&root);
+
+    println!("RUST GENERATOR OWNERSHIP PROBE");
+    println!("  repository: {}", root.display());
+    println!("  generator source: {}", generator.display());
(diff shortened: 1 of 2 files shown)


teamleaderleo (commented):
The evidence decomposition is good, but I’d tighten two extractor assumptions before treating an Oxc pass as evidence for a reusable ownership primitive.

### 1. `fs_call_kind` currently keys only on the last path segment

`foo::write(path, bytes)` and `custom::read_to_string(path)` are currently classified exactly like `fs::write` / `fs::read_to_string` because `fs_call_kind()` checks only `segments.last()`.

For this high-precision research slice, require a recoverable filesystem qualifier (`fs::...`, `std::fs::...`, perhaps an explicitly resolved/imported alias later). Missing direct imports is preferable to manufacturing I/O ownership.

Add a negative control for a project-local `write(...)`/`foo::write(...)` call.

### 2. `.join("literal")` drops the receiver entirely

`literal_join_path_from_call()` currently returns the string argument whenever the method name is `join`. That means something like:

```rust
let path = parts.join("/");
fs::read_to_string(path)?;
```

can be recorded as reading repository path `/` even though the literal is a separator and the receiver carries the real value.

The first slice should require a receiver that is itself recoverably repository-root/path-like, or narrow the accepted syntax to the exact root-building idiom the probe can prove. Again, UNKNOWN/omission is better than a false ownership edge.

### Mechanical state

- generic CI `32216359266` stops at rustfmt;
- research run `32216359280` passes the four probe tests and then stops at Clippy `collapsible_if`, before the pinned Oxc checkout/replay.

Once these precision controls and mechanical checks are clean, the Oxc discriminator is a strong next gate. The intended relation remains well framed: literal generator I/O proves participation, not that every source edit requires regeneration.

teamleaderleo (commented):
Follow-up on the current green head: the filesystem-call qualifier issue is fixed and the Oxc replay/receipt are strong, but the `.join("literal")` receiver issue from my earlier review is still present.

`literal_join_path_from_call()` currently accepts any method call named `join` with one string literal and returns only that literal. The new negative test covers `custom::write`, but not a false path constructor such as:

```rust
let separator = parts.join("/");
std::fs::read_to_string(&separator)?;
```

That still becomes a synthetic repository read of `/`.

For Oxc, the accepted receiver is concretely `root`, and that binding comes from `project_root::get_project_root()`. The probe can stay narrow: either prove a receiver/root origin that this corpus actually supplies, or downgrade unresolved joins instead of treating every string join as a repository path.

A high-precision first option could track bindings whose initializer is an explicitly recognized repository-root provider, then allow `.join(literal)` only on those bindings. It is also fine for the generic probe to miss aliases until another corpus demands them.

The durable research note should keep the successful Oxc relation, but I would not merge the reusable example while this false ownership edge remains. The whole point of this layer is to make explicit ownership stronger than historical correlation.

Issue #62, opened 2026-08-19
Prototype bounded agent context packets before repository edits


teamleaderleo:
## Idea

Cultist already has the ingredients an automated coding agent needs before editing a repository: scoped precedent, historical companions, explicit guidance, project memory, counterexamples, provenance, and explicit unknowns.

Add a **bounded agent context packet** that composes those evidence sources for one target without inventing a separate AI reviewer.

Possible command:

```text
cargo cultist brief path/to/file.rs
cargo cultist brief path/to/item --format json
```

The first consumer is an agent asking:

> What repository evidence would I regret missing before I modify this target?

## Product role

This should be a view over existing evidence primitives, especially:

- #12 — `why` evidence packets;
- #18 — Git / PR / issue / review project memory;
- #38 — explicit repository guidance;
- #4 / #34 — temporal precedent and historical companions;
- #6 — counterexample search;
- #15 / #22 — provenance and machine-readable findings;
- #41 — agentic-history corpus and handoff recovery.

It should not create a parallel taxonomy or require an LLM.

## Agent requirements

- bounded by default, with explicit truncation receipts;
- keep authority/guidance separate from observed frequency;
- include counterexamples before promoting a convention;
- preserve provenance / trust-domain distinctions;
- treat `UNKNOWN` as useful output;
- keep remote PR/issue/review enrichment optional and separately labeled.

## First deterministic slice

For one existing file target, local-only v0 can compose:

1. repository-relative target identity;
2. applicable `AGENTS.md` / `CONTRIBUTING.md` artifacts by ancestor scope;
3. bounded recent non-merge target history;
4. bounded historical companion support, examples, and counterexamples;
5. explicit unknowns and truncation receipts;
6. later, current analyzer findings relevant to the target.

## Active research: draft PR #63

PR #63 contains the research contract, `examples/agent_context_packet.rs`, and the executed replay receipt.

### SmolRunner replay result

Pinned local-only replay against `teamleaderleo/smolrunner@ed3b70e375a57eabce26f2311f798f75b33bdeb0` succeeded.

For `src/disposable_clone_runtime.rs`, the default packet recovered root `AGENTS.md` plus all four intended earned-history discriminators:

```text
#540  Remove duplicate worker readiness probes
#538  Reduce live admission latency before JIT
#533  collapse redundant clone admission polls
#530  keep clone preflight before durable checkpoint
```

[...]

Issue #41, opened 2026-08-18
Use Stensibly agentic churn as an organizational-precedent research corpus


teamleaderleo:
## Idea

`teamleaderleo/stensibly` is a unusually dense longitudinal corpus of agent-produced repository activity: many parallel workers, frequent handoffs, explicit operating-policy revisions, live-provider dogfood, fix-forward repairs, duplicate lanes, invalid proof attempts, stale evidence, and recurring promotion of learned lessons into tests, contracts, CI, and durable state.

Cultist should use that history as a **research feeder for organizational precedent**, analogous to how #29 uses Linux Fieldwork bug species as a feeder for code-level analyzer ideas.

The bridge would be:

```text
agentic episode in Stensibly
-> classify the coordination/failure species
-> identify the repository evidence that exposes it
-> ask whether the pattern is recoverable deterministically from Git/PR/issue/policy history
-> prototype a Cultist evidence packet or finding
-> replay against other agent-heavy repositories
-> keep, weaken, split, or reject the analyzer
```

The goal is not to turn Stensibly's local operating habits into universal rules. Stensibly supplies unusually rich **executed cases and counterexamples** for learning what repository history can say about agent organizations.

## Why this corpus is interesting

Traditional coding-agent benchmarks usually score one task or trajectory in isolation. Stensibly preserves longer causal chains where a locally reasonable change changes the environment for later agents.

That makes it possible to study questions such as:

- did a discovered failure recur after its lesson was recorded?
- did the lesson remain prose, or become a test / invariant / CI gate / server-enforced fact?
- did a process optimization create a later correctness or security regression?
- did parallel workers independently produce overlapping repairs?
- did an agent act on stale, ambiguous, or semantically adjacent evidence?
- did repository policy itself change after repeated agent/operator friction?
- can later workers recover responsibility, authority, evidence, and the next action without the original chat?

### 1. Failure -> generalized executable guard

Stensibly PR #1575 (`Catch overlong Convex index identifiers in CI`) follows production deployment failures where ordinary CI admitted index names that Convex rejected at deploy time. The repair generalizes the provider constraint into a repository-wide regression guard instead of fixing only one identifier.

This is a strong example of **lesson promotion**:

```text
specific failure
-> causal class
-> deterministic invariant
-> future automatic enforcement
```

Possible Cultist question: when the same failure class appears repeatedly, does the repository eventually gain a common guard, or continue accumulating one-off repairs?

### 2. Useful optimization -> second-order control regression

[...]

Issue #18, opened 2026-08-18
Treat Git, PRs, issues, and reviews as project memory


teamleaderleo:
## Idea

Important project rationale often lives outside the final source tree: commit messages, pull-request discussions, review comments, and issues.

Cultist should eventually be able to treat those artifacts as **project memory** while preserving source/provenance and keeping remote access optional.

## Explore

Local-first sources:

- commit messages
- merge commits
- tags/releases
- blame/history

Optional connected sources:

- pull-request descriptions
- inline review threads
- issue discussions
- linked bug reports / design discussions

## Use cases

- recover why an unusual line or exception was introduced
- distinguish a deliberate local exception from an accidental deviation
- attach review rationale to a promoted lint/decision record
- discover that a workaround was introduced for an issue that is now closed/resolved

## Principles

- Source artifacts are evidence, not unquestionable truth.
- Remote project-memory analysis must remain optional; local Git-only mode stays useful.
- Every recovered rationale should identify its origin and date.
- Contradictory historical explanations should be surfaced as disagreement, not silently reconciled.

## Success criteria

- A `why` or archaeology packet can include relevant commit rationale.
- Optional GitHub-backed evidence can enrich the packet without becoming required for normal operation.
- Findings preserve links/identifiers to their source artifacts.

teamleaderleo:
The first provider-backed project-memory loop is now merged through #160 + #162.

Offline boundary (#160): strict bounded packet with exact artifact identity/revisions/paths/text and explicit typed relationships. Normal Cultist commands import none of this and make zero provider calls.

Optional GitHub collector (#162): one selected PR -> recognized explicit same-repository references -> fetch exactly those named artifacts -> emit packet -> independent Rust validation. No broad search, timeline crawl, nearby-PR heuristic, review/comment ingestion, or chronology-derived edges.

Executed Stensibly #1575 receipt:

```text
run:      32243862728
job:      96040065806
artifact: 9361842320
sha256:   2acd2451a01c4b2a48f2f54a2c5de0b06c94c855931ca2e9b2c6b95239769e4e
```

Live packet discriminator:

```text
artifacts: 5
explicit anchor edges: 4
changed path: test/convex-index-identifier-limit.test.ts
closes -> issue #1574
follow_up_to -> PR #1569
follow_up_to -> PR #1571
follow_up_to -> PR #1573
```

[...]

Issue #12, opened 2026-08-18
Add `why` mode with evidence packets and explicit unknowns


teamleaderleo:
## Idea

Add an interactive/explanatory mode that answers repository-level "why might this exist?" questions by assembling evidence first.

Example:

```text
cargo cultist why pci/src/vfio.rs:361
```

Possible output sections:

- DIRECT EVIDENCE — comments, contracts, types, local checks
- CALLER EVIDENCE — how the value/result is consumed
- PRECEDENT — analogous implementations
- HISTORY — introduction/change commits and messages
- COUNTEREVIDENCE — examples that behave differently
- UNKNOWN — rationale that cannot be recovered

Optional LLM interpretation can eventually sit on top of this packet, but packet construction should remain useful and inspectable without a model.

## Explore

Start with a narrow target such as a function, item, or line and gather only bounded relevant context. Avoid sending entire repositories to a model.

Possible Git primitives:

- `git blame`
- `git log -S` / `-G`
- merge-base-aware history
- rename/copy following where practical

## Success criteria

- `why` can produce a useful deterministic evidence packet for at least one real repository example.
- Missing rationale is printed as `UNKNOWN` rather than filled in.
- Optional explanation code is separated cleanly from evidence collection.

PR #63, opened 2026-08-19
research: define bounded agent context packets

teamleaderleo:
## Question

Can Cargo Cultist give a coding agent a small deterministic packet of repository evidence that improves the agent's next action without dumping the repository into a model or inventing missing rationale?

## What this captures

Adds the research contract, a standalone local prototype, and an executed SmolRunner replay for #62.

### `research/agent-context-packet.md`

Defines:

- `cargo cultist brief TARGET` as a possible composition view over existing evidence primitives;
- a bounded machine-readable packet with explicit truncation receipts;
- path-scoped repository guidance as first-class evidence;
- counterevidence and `UNKNOWN` as required agent inputs;
- source/origin provenance for trust-domain distinctions;
- SmolRunner replay targets where source-local inspection misses earned history;
- a Stensibly negative control where temporal adjacency must remain distinct from causality;
- evaluation based on agent decisions/evidence accesses instead of one aggregate score.

### `examples/agent_context_packet.rs`

The first local-only probe emits deterministic JSON for one file target containing:

- repository-relative target identity;
- applicable ancestor-scoped `AGENTS.md` / `CONTRIBUTING.md` artifacts;
- up to 20 recent non-merge commits touching the target;
- up to 8 historical companion paths with support, examples, and counterexamples;
- explicit exclusions/truncation receipts;
- explicit `UNKNOWN`s for remote rationale, guidance prose interpretation, causal inference, and analyzer composition.

### Executed replay

`research/agent-context-packet-replay.md` records an exact pinned replay against SmolRunner.

The local packet for `src/disposable_clone_runtime.rs` recovered all four intended earned-history discriminators within the default 20-commit budget:

```text
#540  Remove duplicate worker readiness probes
#538  Reduce live admission latency before JIT
#533  collapse redundant clone admission polls
#530  keep clone preflight before durable checkpoint
```

It also found root `AGENTS.md`, counterexample-bearing historical companions, and the likely adjacent surfaces (`DISPOSABLE_AUTOSCALING_CI.md`, worker/coordinator code, the durable clone transaction, and prepared-template files).

The packet for `src/host_preparation_command.rs` recovered the exact local repair commit (`Bind host-preparation confirmation to exact durable plan`) and a 2/2 companion relation with its test file.

The replay also exposed a useful boundary: local Git recovers the repair but does not contain all of PR #555's deeper privilege/authorization rationale. That supports #18 optional PR/issue/review enrichment as a separate evidence layer instead of pretending local commit text is equivalent to remote project memory.

Temporary external replay CI was removed after the receipt was captured. Normal repository CI remains the ongoing check.

## Boundary

Research only. This PR does not add an AI reviewer, LLM dependency, automatic rule promotion, or public CLI command.

Refs #62, #12, #18, #38, #41.


teamleaderleo (commented):
The local packet probe is a good research slice, and the SmolRunner discriminators are pointed at the right thing: recover the earned recent-history subjects and applicable guidance artifact without interpreting prose or manufacturing causality.

Two contract gaps are worth making explicit before calling the packet itself “bounded evidence” in the stronger sense used by the design note.

### 1. Per-relation example truncation is currently silent

`historical_companions()` does:

```rust
present_in.iter().take(max_examples_per_relation)
...
counterexamples.take(max_examples_per_relation)
```

but the top-level `truncation` array only reports omitted companion paths, excluded commits, and the recent-history limit. A consumer cannot tell whether a relation had 2 examples total or 40 with 38 omitted.

The note correctly says silent truncation is unacceptable for an agent-facing interface. Add a relation-level count (`examples_omitted`, `counterexamples_omitted`) or a top-level receipt that preserves this fact.

### 2. v0 is count-bounded, not byte-bounded yet

The design and evaluation metric talk about `max_bytes` / useful evidence per packet byte. The current `PacketBudget` has history/companion/example/path-count bounds only. That is completely reasonable for a research prototype, but keep the claim precise: this head demonstrates deterministic **count-bounded** selection, not yet a hard serialized-size budget.

I would either record `serialized_bytes` in the research output now (cheap measurement) or explicitly leave byte budgeting as an `UNKNOWN`/next experiment. Before a public `brief` command, implement a deterministic byte budget with receipts so a long commit subject/path cannot blow through the intended context envelope.

### What I like

- ancestor-scoped guidance discovery stays source-only;
- recent history and co-change are separate sections;
- `UNKNOWN` names missing remote rationale / prose interpretation / causality;
- the SmolRunner workflow checks specific earned-history subjects, so the replay is testing selection quality instead of JSON validity alone.

Normal CI on the current head is green. The dedicated replay should remain a research gate; once its receipt is captured, retire the temporary external workflow the same way #39/#54/#59 have been doing.

Issue #65, opened 2026-08-19
Prove repository-root origin in Rust generator ownership paths


teamleaderleo:
## Problem

Merged PR #59 successfully recovered the intended Oxc generator relation, but its reusable research probe still accepts a broader syntax class than it can justify.

`literal_join_path_from_call()` currently treats any method call of the form:

```rust
receiver.join("literal")
```

as a repository-relative path and returns only the literal argument. The receiver is discarded.

That can manufacture ownership edges. For example:

```rust
let separator = parts.join("/");
std::fs::read_to_string(&separator)?;
```

can be recorded as a repository read of `/` even though the join is string/collection assembly rather than filesystem path construction.

The Oxc acceptance case is real and narrower:

```rust
let root = project_root::get_project_root()?;
fs::read_to_string(root.join("crates/oxc_linter/src/rules.rs"))?;
```

## Desired evidence contract

A literal suffix should become a repository-relative path only when the probe can recover enough receiver provenance to justify that interpretation.

A good first slice can stay narrow. Options include:

- recognize an explicit repository-root provider and track the resulting binding;
- track a small typed/path constructor idiom when its origin is machine-visible;
- otherwise omit the edge / classify receiver origin as `UNKNOWN`.

Do not use variable names such as `root` alone as proof.

## Regression controls

Positive:

```rust
let repo = recognized_repository_root_provider()?;
std::fs::read_to_string(repo.join("src/rules.rs"))?;
```

Negative:

```rust
let separator = parts.join("/");
std::fs::read_to_string(&separator)?;
```

Negative:

```rust
let display = names.join("generated/output.rs");
std::fs::write(display, bytes)?;
```

The latter two must not become repository ownership edges.

## Why keep this separate from #59

#59's executed Oxc receipt remains useful: for that pinned source, the receiver really is the repository root and the exact read/write relations are present. This follow-up tightens the generic adapter boundary before it is composed into a missing-generated-companion finding.

Refs #59, #35, #15.

PR #66, opened 2026-08-19
experiment: compose generated-companion diff evidence

teamleaderleo:
Superseded by #68 after the stronger held-out replay passed.

This branch proved the evidence layers could be composed on a synthetic Oxc registration edit.

Retained receipt:

```text
head:     a561ab583f9980183f4989ac0d27534e8a25ffbf
workflow: Generated companion diff packet research
run:      32217441904
job:      95961477574
result:   success
artifact: 9352810946
sha256:   a56dd3ba451f9dbd4c9f4a4be16b85c530f19463abf892ed8f41a9e114d8f173
```

The synthetic packet recovered current diff absence, `cargo lintgen`, literal generator data movement, generated markers, `.gitattributes` declarations, and 99/99 directional syntax-cohort support.

#68 then repeated the composition against a genuinely held-out real Oxc target and the exact historical docs-only exception:

- real target's source-only counterfactual -> bounded finding;
- docs/license-only historical exception -> `NO FINDING`.

#68 is therefore the canonical proof specimen and product-extraction source. No unique product code remains here.

### DIFF
--- a/.github/workflows/generated-companion-diff-packet-research.yml
+++ b/.github/workflows/generated-companion-diff-packet-research.yml
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       - experiment/generated-companion-diff-packet
+  pull_request:
+    paths:
+      - .github/workflows/generated-companion-diff-packet-research.yml
   workflow_dispatch:
 
 permissions:
@@ -100,6 +103,7 @@ jobs:
           python - <<'PY'
           import json
           import re
+          import subprocess
           from pathlib import Path
 
           anchor = 'crates/oxc_linter/src/rules.rs'
@@ -110,7 +114,7 @@ jobs:
 
           changed = set(
               line.strip()
-              for line in __import__('subprocess').check_output(
+              for line in subprocess.check_output(
                   ['git', '-C', 'corpus/oxc', 'diff', '--name-only'], text=True
               ).splitlines()
               if line.strip()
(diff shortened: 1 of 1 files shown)


teamleaderleo (commented):
This is the right composition experiment. The workflow keeps current change, explicit guidance, generated identity, ownership, historical cohort, and negative space as separate inputs, then asks one bounded question.

One addition before I trust the ownership layer in this replay: explicitly pin the **repository-root receiver provenance** for Oxc.

Issue #65 exists because the merged generic `rust_generator_ownership` research example still treats any `receiver.join("literal")` as a repository-relative path. For this exact Oxc source, we independently know the relevant receiver is real:

```rust
let root = project_root::get_project_root()?;
fs::read_to_string(root.join("crates/oxc_linter/src/rules.rs"))?;
```

Add an exact source assertion in the pinned workflow (or another narrow deterministic check) that the generator function's `root` binding comes from `project_root::get_project_root()` before setting `generator_reads_anchor_and_writes_output = true` in the composed packet.

That lets this controlled Oxc packet remain valid even while #65 tightens the generic adapter.

Two smaller notes:

- After #59 merged, retarget this PR to `main` before final disposition so the research packet is evaluated against the actual integrated adapter.
- If the pinned replay passes, record the exact run/artifact and remove this temporary external workflow before merge, as with #39/#54/#58/#59.

I like the evidence boundary in the human packet: ownership + generated declarations + 99/99 syntax precedent still stop at a question, and the packet explicitly says every possible anchor edit need not change generated bytes.

PR #68, opened 2026-08-19
experiment: compose generated-companion evidence into a held-out diff finding

teamleaderleo:
Proof specimen retained; product extraction landed in #72.

## What this proved

Held-out Oxc target:

```text
```

The source-only counterfactual emitted the bounded generated-companion finding using parent-only history, while the exact historical docs/license exception `5e113baf...` stayed silent.

Exact proof receipt:

```text
head:     746d0ab4e6786099baa618563653c8ddc62c4660
generic:  32217601800 — success
research: 32217601874 / job 95961927494 — success
artifact: 9352863272
sha256:   faac0d7d5cd66f1f28353562b2b1ccbcda1577be031ecd1dfc03d118f93f258d
```

## Product promotion

PR #72 extracted the behavior into normal `cargo cultist diff`, caught and repaired the removal-only diff plumbing bug, replayed the same held-out positive + docs-only negative control through the real product binary, rebased onto current main, passed exact-head CI, and merged as:

```text
1968c5774091f19630004f3241e1d87fa4a4965a
```

The long research example remains useful history on this branch; the durable product receipt now lives on main in `research/generated-companion-product-replay.md`. No unique landing responsibility remains here.


teamleaderleo (commented):
The held-out design is substantially stronger than the synthetic composition cell in #66. Resetting to the target's first parent and applying only the source patch means the analyzer cannot use the answer commit in its historical cohort, and replaying `5e113baf...` as a docs-only negative control is exactly the right falsification.

One method gate before the result is trustworthy: explicitly prove the pinned Oxc generator's repository-root receiver provenance.

This example currently copies the unresolved #65 behavior:

```rust
receiver.join("literal") -> repository path literal
```

without interpreting the receiver. For Oxc we know the intended edge is narrower:

```rust
let root = project_root::get_project_root()?;
fs::read_to_string(root.join("crates/oxc_linter/src/rules.rs"))?;
```

Add an exact pinned-source assertion (or another narrow deterministic receiver-origin check) before accepting generator ownership in the held-out workflow. Then the **Oxc** result remains valid while #65 separately repairs the generic adapter.

Current research run `32217435381` has not reached Oxc yet; it stops at rustfmt in the probe quality gate. The temporary format workflow can handle that mechanically, but please remove any write-enabled formatting workflow before final research disposition.

If the held-out positive and exact docs-only negative both pass after the root-provenance check, I would treat #68 as the stronger composition result and likely close/supersede #66 rather than retaining two workflow-glue experiments. The durable record should explain why the held-out target was unavailable to the cohort and preserve the exact negative-control outcome.

PR #71, opened 2026-08-19
experiment: compare lexical lock acquisition with declared rank policy

teamleaderleo:
## Question

Can Cargo Cultist recover a repository-declared lock-order policy and identify a real function whose lexical acquisition order contradicts it?

This is the first relational-precedent experiment aimed directly at concurrency correctness.

## WGPU discriminator

Historical Fieldwork work isolated `Queue::compact_blas_inner` on exact WGPU trunk:

`gfx-rs/wgpu@422b1c3a2c08feb39e63fdb7ca2798b26803d427`

The repository itself declares a lock-rank DAG in `wgpu-core/src/lock/rank.rs`:

```text
DEVICE_COMMAND_INDICES -> QUEUE_PENDING_WRITES
```

while `QUEUE_PENDING_WRITES` does not allow `DEVICE_COMMAND_INDICES` as a follower.

The trunk function acquires `pending_writes`, keeps it held, explicitly drops the snatch guard, then acquires `command_indices`.

Public repair control:

`andyleiserson/wgpu@7350b0c8d0acb2d410e3f5922fd4aba8407d5cbc`

That version acquires `command_indices` before `pending_writes` and keeps the snatch guard.

## Probe

`examples/lock_order_policy.rs` intentionally handles a narrow lexical subset:

- parses the repository's `define_lock_ranks!` invocation;
- maps unique rank member field names such as `Device::command_indices` and `Queue::pending_writes`;
- finds one named impl function;
- tracks top-level named guards initialized by `.lock()`, `.read()`, or `.write()`;
- treats explicit `drop(guard)` as release;
- checks each acquisition made while another supported guard remains held against the declared follower set.

Nested control flow, helper-returned guards, aliases, temporary guards, and ambiguous field names remain outside this first probe.

## Exact controls

The read-only research workflow must:

1. emit a rank-policy contradiction on exact WGPU trunk for `pending_writes -> command_indices`;
2. stay quiet on the exact public repair head;
3. show the repaired `command_indices -> pending_writes` acquisitions explicitly.

This experiment uses explicit repository policy as the authority. Historical acquisition mining can be added later as corroborating precedent instead of replacing the declared DAG.


teamleaderleo (commented):
The model matches WGPU's declared policy more closely than a generic lock-order heuristic would. `rank.rs` explicitly says the rank of the **most recently acquired lock still held** determines permitted next acquisitions, which is exactly what this probe tracks for its supported top-level named guards.

The lexical boundary is also appropriately narrow for this first corpus: direct `.lock()` / `.read()` / `.write()` on uniquely ranked fields plus explicit `drop(guard)`, with nested/control-flow/alias cases left out.

Current research run `32218155334` never reaches WGPU because `cargo fmt --check` stops on the `find_impl_function` signature; generic CI is red for the same mechanical state.

My acceptance bar after formatting:

1. exact WGPU trunk `422b1c3a...` emits the declared-rank contradiction for `pending_writes -> command_indices`;
2. exact repair head `7350b0c8...` stays clean and visibly acquires `command_indices -> pending_writes`;
3. record the exact run/artifact and the repository policy quotation in a durable research note;
4. remove the temporary external workflow before merge.

If that matrix passes, this is worth keeping as a new relation species: **explicit repository policy contradicted by a supported local operation sequence**. Historical lock-order mining can remain corroborating evidence later; it should not outrank the repository's own rank DAG in this case.

PR #72, opened 2026-08-19
Detect missing generated companions in diffs

teamleaderleo:
Product extraction from the successful held-out Oxc proof in #68.

## Product boundary

`cargo cultist diff` now emits a high-precision `generated-companion-missing` finding only when all of these independent repository facts agree:

- the current diff changes normalized Rust syntax in a source file;
- a Cargo alias resolves to `cargo run -p PACKAGE`;
- that generator package's default Rust entrypoint contains a literal same-function repository-path read from the changed source and write to an output;
- the output exists, self-identifies with a strong generated header, and the exact path is marked `linguist-generated=true` in `.gitattributes`;
- the output is absent from the current diff;
- at least three comparable source syntax-changing commits exist and the output changed in every one of them.

The finding keeps evidence layers separate and ends with `UNKNOWN` current intent / whether this particular source edit changes generated bytes.

## Held-out positive

Pinned Oxc target:

```text
```

First parent:

```text
568203e24f090f2cc3f945d611e605b864842bf0
```

The actual target independently changed:

```text
crates/oxc_linter/src/rules.rs
crates/oxc_linter/src/generated/rule_runner_impls.rs
crates/oxc_linter/src/generated/rules_enum.rs
```

The replay withheld the two generated files: it saved only the parent->target `rules.rs` patch, reset to the target parent, and applied the source patch alone. The target commit was therefore absent from the history used to learn precedent.

Product replay run `32218582180`, job `95964645621`, completed successfully. The real `cargo-cultist diff --format json` command emitted exactly two findings, one per missing generated registry.

Each finding carried:

```text
derived   current normalized Rust syntax changed; output absent
derived   cargo lintgen / oxc_linter_codegen literal source->output ownership
observed  generated header + exact .gitattributes declaration
observed  99/99 comparable Rust syntax-changing commits
unknown   current intent / whether this source edit changes emitted bytes
```

## Exact historical negative control

The same product replay then used:

```text
5e113baf716b9f3781331b268b4142d23cac0541
docs(linter): add license notices for ported ESLint plugins (#22768)
```

Again only the `rules.rs` patch was applied to its first parent and every companion was withheld.

Normalized Rust syntax was unchanged after comments/docs/whitespace were removed, and the product emitted zero `generated-companion-missing` findings. The workflow failed closed if any appeared.

## Extraction bug caught by the held-out replay

The first product build returned zero findings on the held-out target even though the independent generator-ownership and 99/99 cohort probes were green.

Root cause: the real Oxc source patch is removal-only. The pre-existing test-module diff analyzer intentionally defined its Rust worklist from **added lines** and returned early when that set was empty.

The repair preserves two distinct diff facts:

```text
changed path
  -> generated negative-space analysis

file with parsable added lines
  -> test-module naming analysis
```

Generated-companion analysis now runs before the added-line early return. The original module-naming analyzer keeps its existing semantics.

## Validation / receipts

Held-out product proof:

```text
run 32218582180 / job 95964645621 — success
```

Pre-rebase clean generic CI:

```text
run 32218582111 / job 95964645017 — success
```

The temporary held-out network workflow was removed after the receipt. Full methodology and boundaries are retained in `research/generated-companion-product-replay.md`.

The branch was then compacted using GitHub's conflict-free merge tree onto current main `ca63aaf411b33aff082811dc65ea120e782328ef`, preserving newly landed lock-order work. Final branch head:

```text
c59f6e4b5a6f6d27adf89e31bd5a074839c054f9
```

Exact-head generic CI run `32218865371` passed every substantive step: rustfmt, Clippy with warnings denied, full tests, repository/history/CI-test dogfood, and diff text + JSON dogfood.

## Deliberate limits

The first slice stays quiet for:

- non-Cargo aliases / aliases that do not map to `run -p`;
- generator packages without default Rust `src/main.rs`;
- dynamic/helper-crossing path construction;
- non-Rust generators;
- generated files lacking both a strong header and exact `.gitattributes` declaration;
- histories with any semantic-cohort counterexample;
- comments/docs/whitespace-only source edits.

These are precision choices, not claims that other generation relationships are unimportant.

## Follow-up

The first extraction is intentionally self-contained and larger than the desired long-term module. Land the proven bounded behavior first, then consolidate repeated Rust/history/generator fact extraction behind shared adapters while keeping this held-out matrix fixed.

Refs #35, #38, #68.

### DIFF
--- a/research/generated-companion-product-replay.md
+++ b/research/generated-companion-product-replay.md
@@ -1,0 +1,207 @@
+# Generated-companion product replay: held-out Oxc diff
+
+Date: 2026-08-19
+
+Status: successful product extraction of the held-out generated-companion proof into `cargo cultist diff`.
+
+## Product question
+
+Can the normal diff analyzer emit a bounded missing-generation finding by composing independently recovered repository evidence, while staying quiet for the exact historical docs-only exception?
+
+The first product slice requires all of these facts:
+
+```text
+current source path changed
++ normalized Rust syntax changed
++ Cargo alias names a generator package
++ generator entrypoint literally reads source and writes output
++ output self-identifies as generated
++ exact output path is linguist-generated=true
++ output is absent from the current diff
++ output changed in every comparable source-syntax history commit
+= provenance-bearing question
+```
+
+Intent and output-byte effect remain `UNKNOWN`.
+
+## Held-out target
+
+Oxc target commit:
+
+```text
+228e8e0f85c0e7aeded02c5e27fd810004d3b41a
+fix(linter): resolve inactive React compiler rules (#25830)
+```
+
+First parent:
+
+```text
+568203e24f090f2cc3f945d611e605b864842bf0
+```
+
+The actual target independently changed:
+
+```text
+crates/oxc_linter/src/rules.rs
+crates/oxc_linter/src/generated/rule_runner_impls.rs
+crates/oxc_linter/src/generated/rules_enum.rs
+```
+
+The replay saved only the target-parent -> target `rules.rs` patch, reset to the parent, and applied the source patch alone. The target commit was therefore absent from the history used to evaluate precedent, and both generated outputs were deliberately absent from the worktree diff.
+
+## First extraction failure: removal-only diff
+
+The initial product build passed rustfmt, Clippy, unit tests, and ordinary Cargo Cultist dogfood, but returned zero generated-companion findings on the held-out target.
+
+A diagnostic replay on the same counterfactual proved the independent evidence adapters were still healthy:
+
+```text
+cargo lintgen -> run -p oxc_linter_codegen
+
+generate_rule_runner_impls
+  reads  crates/oxc_linter/src/rules.rs
+  writes crates/oxc_linter/src/generated/rule_runner_impls.rs
+
+generate_rules_enum_file
+  reads  crates/oxc_linter/src/rules.rs
+  writes crates/oxc_linter/src/generated/rules_enum.rs
+
+semantic cohort:
+  rule_runner_impls.rs  99/99  100.0%
+  rules_enum.rs         99/99  100.0%
+```
+
+The loss was in diff plumbing. The held-out `rules.rs` patch is removal-only. The pre-existing test-module analyzer intentionally defined its Rust worklist from **added lines**, so `build_diff_analysis_report` reached an early return before the generated-companion analyzer ran.
+
+The repair separated the two facts:
+
+```text
+changed path
+  -> generated negative-space analysis
+
+file with parsable added lines
+  -> test-module naming analysis
+```
+
+Generated-companion analysis now runs before the added-line early return. The legacy test-module analyzer keeps its existing added-line semantics.
+
+## Clean product receipt
+
+Clean product head after that repair and removal of the temporary write-capable wiring workflow:
+
+```text
+50f50fbad9c299f8156263670fe86c06dd3b036b
+```
+
+Generic CI:
+
+```text
+run:    32218582111
+job:    95964645017
+result: success
+```
+
+That run passed:
+
+- rustfmt;
+- Clippy with warnings denied;
+- all unit tests;
+- repository text + JSON dogfood;
+- history text + JSON dogfood;
+- `ci-tests` text + JSON + stale/matching fixture;
+- diff text + JSON dogfood.
+
+So the path-level generated analysis did not disturb the original added-line analyzer.
+
+## Held-out product replay
+
+Read-only product replay:
+
+```text
+run:    32218582180
+job:    95964645621
+result: success
+```
+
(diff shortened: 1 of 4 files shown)


teamleaderleo (commented):
The first-product threshold is appropriately conservative, but this branch has not yet achieved the extraction boundary #68 earned.

### 1. Blocker: the product module copies #65's false ownership edge

`src/generated_diff.rs` still has the old logic:

```rust
Expr::MethodCall(call) => literal_join_path_from_call(call)
...
fn literal_join_path_from_call(call: &ExprMethodCall) -> Option<String> {
    // accepts any receiver.join("literal")
}
```

So `parts.join("/")` or an arbitrary `PathBuf` parameter can still become repository ownership in the **product** finding path.

PR #70 is the accepted generic repair: only joins rooted in a binding derived from an explicitly recognized repository-root provider qualify, with adversarial controls and a pinned Oxc replay. #72 should consume that stricter behavior (ideally through one shared helper) before any held-out replay is considered meaningful.

### 2. The "small reusable analyzer" is currently 910 lines

The held-out research carrier was intentionally large because it had to prove the entire composition. `src/generated_diff.rs` currently re-copies substantial pieces of:

- generator ownership extraction;
- generated-marker parsing;
- Cargo alias/package discovery;
- Git history/cohort logic;
- Rust syntax fingerprinting.

That is very close to promoting the research carrier under a new filename. Before product integration, extract/share the primitives whose evidence contracts already exist. In particular, avoid creating two independent versions of generator-path provenance after #70 and two independent versions of syntax/history semantics that can drift from `history` / shared Rust facts.

I don't require a grand refactor in this PR, but the ownership extractor must have one canonical implementation and the diff finding should be visibly smaller than the proof specimen.

### 3. Remove the write-enabled wiring workflow

`.github/workflows/wire-generated-companion-diff.yml` has `contents: write` and auto-commits/pushes source changes. It has already done its job. Delete it before the validation phase; the final product PR should be ordinary source + CI, with any pinned external replay read-only and temporary.

### 4. Keep the threshold exactly this strict for v1

I like these gates for the first product slice:

```text
changed normalized Rust syntax
+ proven generator ownership
+ exact generated declaration/header
+ output absent from diff
+ >=3 comparable syntax-changing commits
+ 100% directional support
-> bounded finding with UNKNOWN intent
```

Do not relax `100%` or the minimum cohort while extracting. The first product feature should inherit the evidence level of the held-out witness; broader support ratios belong in later evaluation.

### Acceptance

After #70 semantics are integrated and the module is reduced/shared:

1. generic CI green;
2. replay the exact #68 held-out positive through the real `cargo cultist diff` command;
3. replay the exact docs/license negative and require no generated-companion finding;
4. record the receipt and remove the temporary external replay workflow.

That would be a real promotion from research to product, rather than a copy of the research implementation.

teamleaderleo (commented):
The held-out replay found the primary product-integration bug, and it is more important than the size/refactor comments.

### Blocker: deletion-only Rust changes disappear before generated analysis

Research run `32218308776` reaches the held-out Oxc counterfactual, but the real `cargo cultist diff` command prints:

```text
DERIVED
  The diff contains added lines in 0 Rust file(s)...
OBSERVED
  No added or renamed test-gated module declarations were found in the diff.
```

and exits before any generated-companion finding. The workflow then fails its `grep` for `generated-companion-missing`.

That target (`228e8e0f...`) is exactly the #68 held-out witness: its `rules.rs` source change is deletion-only. The research carrier classified it via before/after normalized Rust syntax, while the public diff pipeline currently derives `changed_rust_paths` only from added-line ranges:

```rust
let changed = git_diff_changed_lines(...);
let changed_rust_paths = changed.rust_paths();
if changed_rust_paths.is_empty() {
    return Ok(analysis);
}
add_generated_companion_findings(...);
```

So the product integration has accidentally inherited the test-module analyzer's **added-line** relevance model.

Generated-companion analysis needs **changed-file identity**, including deletion-only changes. Test-module precedent still needs added-line ranges. Keep those facts separate.

A good fix is to make the diff scan retain both:

```text
changed paths (A/M/D/R as relevant)
added line ranges by path
```

Then:

- run generated-companion analysis when a Rust source path changed, even if it has zero added lines;
- keep test-module declaration matching scoped to added ranges;
- make the top-level claims precise: `N Rust files changed`, separately from `M Rust files contain added lines` if both are useful.

Add a regression where one Rust file has only deleted syntax and the generated analyzer is still invoked.

### Existing blockers remain

- `src/generated_diff.rs` at this head still accepts arbitrary `receiver.join("literal")`; #70 is merged on `main` and must become the canonical ownership behavior before product promotion.
- remove the write-enabled wiring workflow before final validation.
- reduce/share the copied research logic rather than making the 910-line module the long-term product boundary.

This red held-out replay is a very good result: it found a genuine abstraction mismatch between two analyzers before merge. Fix the change-vs-added-lines model first, then rerun the exact #68 positive and docs-only negative through the public command.

PR #77, opened 2026-08-19
experiment: normalize numeric lock-rank policy on asupersync

teamleaderleo:
Superseded by #79.

The first adapter was useful failure evidence: it exposed two target assumptions before any corpus claim could pass—incorrect inventory-schema assumptions and source extraction that was too shallow for the repository's oracle. #79 keeps the same normalized numeric policy while rebuilding the source adapter cleanly from those lessons.

### DIFF
--- a/.github/workflows/numeric-lock-order-policy-research.yml
+++ b/.github/workflows/numeric-lock-order-policy-research.yml
@@ -1,0 +1,83 @@
+name: Numeric lock order policy research
+
+on:
+  push:
+    branches:
+      - experiment/numeric-lock-order-policy
+  pull_request:
+    branches:
+      - main
+
+permissions:
+  contents: read
+
+jobs:
+  asupersync:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout cargo-cultist
+        uses: actions/checkout@v4
+        with:
+          path: cultist
+          persist-credentials: false
+
+      - name: Install stable Rust
+        run: |
+          rustup toolchain install stable --profile minimal --component rustfmt,clippy
+          rustup default stable
+
+      - name: Probe quality gates
+        working-directory: cultist
+        run: |
+          cargo fmt --check
+          cargo clippy --example numeric_lock_order_policy -- -D warnings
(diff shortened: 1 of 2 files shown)


teamleaderleo (commented):
This is exactly the second corpus #71 needed, but the current adapter is parsing a JSON schema the pinned repository does not have.

### Blocker: inventory field shapes do not match asupersync

`parse_exact_name_ranks()` currently expects:

```text
rank_map[].rank                 -> number
rank_map[].from_name_samples    -> object{name: rank}
```

The pinned `artifacts/lock_order_inventory_v1.json` actually contains:

```json
{
  "rank_name": "Config",
  "numeric_rank": 10,
  "from_name_samples": ["config_cache"]
}
```

(and analogous entries for Regions=30, Tasks=40, etc.). So once the probe gets past formatting it will fail before producing the intended policy map.

Parse `numeric_rank` and the array of exact sample names. The inventory already ties each entry to `source_file: src/sync/lock_ordering.rs`, and the runtime source explicitly documents: lower numeric values are acquired before higher values. The repository-owned tests provide the positive/negative oracle:

```text
test_correct_lock_ordering:
  config_cache -> regions_table -> tasks_queue

test_lock_ordering_violation:
  tasks_queue -> config_cache
  #[should_panic(expected = "Lock ordering violation")]
```

### Current mechanical state

Research run `32218871032` stops at rustfmt before checking out asupersync; generic CI is red for the same formatting state.

### Acceptance after parser + formatting repair

1. require the exact pinned inventory contract/version and parse `numeric_rank` + string sample array;
2. keep `Config 10 -> Regions 30 -> Tasks 40` quiet;
3. detect `Tasks 40 -> Config 10` as descending while both guards remain in scope;
4. record the exact run/artifact and explicit source-policy evidence in a durable research note;
5. retire the temporary external workflow before merge.

If that passes, this is strong evidence that WGPU's successor-DAG policy and asupersync's numeric-rank policy can normalize to one conceptual predicate while keeping repository-specific adapters separate.

Issue #80, opened 2026-08-19
Use proven repository-root provenance in generated diff analyzer


teamleaderleo:
## Product correctness debt

PR #72 merged the held-out generated-companion detector to `main`, but `src/generated_diff.rs` still contains the pre-#70 generic ownership inference:

```rust
receiver.join("literal") -> repository-relative path
```

The receiver is discarded. This is the same false-ownership class fixed in the research adapter by #70 / #65.

Current `main` can therefore manufacture product ownership edges from code such as:

```rust
let separator = parts.join("/");
std::fs::read_to_string(&separator)?;

let display = names.join("generated/output.rs");
std::fs::write(display, bytes)?;
```

If the other strict product gates happen to align, those synthetic paths can participate in a `generated-companion-missing` finding.

The Oxc held-out witness remains valid: its generator really obtains `root` from `project_root::get_project_root()` and uses that root for the exact source/output paths. The problem is the generic product adapter boundary.

## Required repair

Product generator ownership must use the proven-root semantics already demonstrated by #70:

- recognize a repository-root binding only from an explicit supported provider;
- accept literal `.join(...)` paths only when the receiver resolves to that proven root binding;
- preserve bound path variables derived from a proven root;
- omit unresolved receiver origins instead of inferring from variable names or literal suffixes;
- keep filesystem-call qualification independent.

Prefer one canonical ownership implementation shared by the research adapter and product analyzer so these semantics cannot drift again.

## Regression controls

Positive:

```rust
let repo = project_root::get_project_root()?;
let input = std::fs::read_to_string(repo.join("src/rules.rs"))?;
let output = repo.join("generated/rules.rs");
std::fs::write(&output, input)?;
```

Negative:

```rust
fn generate(root: PathBuf) { ... root.join("src/rules.rs") ... }
```

A parameter merely named `root` is insufficient.

Negative:

```rust
parts.join("/")
names.join("generated/output.rs")
```

These must produce no repository ownership edges.

## Acceptance

1. product unit/regression controls above pass;
2. exact #68 held-out Oxc positive still emits two findings through public `cargo cultist diff`;
3. exact docs/license negative remains quiet;
4. generic CI green;
5. no duplicate unsafe ownership implementation remains in `src/generated_diff.rs`.

Treat this as a product-correctness blocker before relaxing generated-companion thresholds or adding new generator adapters.

Refs #72, #70, #65, #68.

teamleaderleo:
[...]

PR #81, opened 2026-08-19
Retire superseded generated-companion proof example

teamleaderleo:
Cleanup after #72 landed the held-out generated-companion detector in `cargo cultist diff`.

The deleted `examples/generated_missing_companion.rs` was the broad proof carrier used by #68. Its unique evidence has moved to durable research receipts, and the production behavior now lives in `src/generated_diff.rs` with a successful held-out product replay.

This PR deletes only that superseded executable specimen. It keeps:

- `research/generated-missing-companion-held-out-replay.md` — held-out proof methodology;
- `research/generated-companion-product-replay.md` — product extraction, removal-only diff bug, positive + docs-only receipts;
- the smaller reusable `rust_syntax_cohort` and `rust_generator_ownership` research tools;
- the merged product analyzer.

No product semantics change. Full CI should prove all remaining examples/tests compile without the retired carrier.


teamleaderleo (commented):
Hold this cleanup until #80 is resolved.

The broad #68 proof carrier is superseded as an implementation, but it is still the easiest executable reference for the exact provenance semantics that the shipped product analyzer currently violates: `src/generated_diff.rs` on `main` still accepts arbitrary `receiver.join("literal")` as repository ownership, while #70/#68 prove the stricter repository-root requirement.

Once #80 lands and the public `cargo cultist diff` path has:

- proven repository-root ownership semantics;
- adversarial collection/string-join negatives;
- the exact held-out Oxc positive;
- the exact docs-only negative;

then deleting `examples/generated_missing_companion.rs` is a good cleanup. Until then, retaining the proof carrier gives the repair lane a direct semantic oracle beyond the prose receipts.

This is sequencing only; I agree the carrier should disappear after product equivalence is restored.

Issue #74, opened 2026-08-19
Compose an agent edit lifecycle: brief -> diff -> teach


teamleaderleo:
## Idea

Cultist's agent value is larger than a pre-edit packet. The existing roadmap already contains the pieces of a full repository-memory loop:

```text
BEFORE CODE
cargo cultist brief TARGET
-> recover the bounded repository evidence worth seeing before an edit

DURING CODE
cargo cultist diff ...
-> compare the live change with scoped precedent, companions, counterexamples, guidance, and unknowns

AFTER CODE
cargo cultist teach <finding/decision> ...
-> preserve the rationale for an intentional exception or newly learned rule
-> later promote repeated stable consensus into deterministic policy

next agent
-> brief includes the decision / earned rule
```

The product question is simple:

> Can an agent enter a repository, recover what it needs, work with live evidence, and leave the repository more legible for the next agent?

This should be composition over existing capabilities rather than a separate AI subsystem.

## Existing owners

- **Before:** #62 bounded agent context packets, #12 `why`, #18 project memory, #38 explicit guidance, #34 history.
- **During:** current `diff` analysis, scoped precedent, counterexample-first findings, claim provenance.
- **After:** #10 explicit decision records / `teach`, #11 promotion of repeated consensus into deterministic rules.
- **Cross-cutting:** #15 claim provenance (completed), #41 agentic-history corpus.

## Before code: orient

Input:

```text
intent + target
```

Output should be a bounded packet containing the evidence an agent is most likely to regret missing:

- applicable guidance and scope;
- recent target history;
- known decisions/exceptions;
- historical companions + counterexamples;
- current relevant findings;
- optional PR/issue/review rationale;
- explicit unknowns and truncation receipts.

The packet informs the next inspection. It does not decide the patch.

## During code: reconcile the live change

The live diff should become the natural incremental query surface.

Useful questions:

- Which new/changed facts conflict with local or repository-wide precedent?
- Which historical companions are absent from this diff, and what counterexamples exist?
- Does explicit guidance disagree with observed frequency?
- Did the change cross a trust/authority boundary whose provenance is relevant?
- Did the agent invalidate evidence it relied on earlier in the task?
- Which claims remain `UNKNOWN` and deserve another inspection before continuing?

This can initially remain explicit agent calls rather than a background watcher:

```text
edit
-> cargo cultist diff --format json
-> inspect relevant evidence
-> edit
-> repeat
```

## After code: preserve the lesson

[...]

PR #90, opened 2026-08-19
Restore generated findings on canonical root provenance

teamleaderleo:
## Result

**#80 acceptance is green, including an adversarial review-dogfood case discovered after the first successful held-out replay.**

This PR restores `generated-companion-missing` to public `cargo cultist diff` with one canonical generator-ownership implementation.

## One canonical ownership implementation

Generator ownership now lives in:

```text
src/generator_ownership.rs
```

Both consumers compile that exact file:

```text
examples/rust_generator_ownership.rs
  -> research view

src/generated_diff.rs
  -> product consumer of canonical GeneratorRelation values
```

`src/generated_diff.rs` no longer contains a separate repository-path ownership parser.

The shared implementation owns:

- explicit repository-root provider recognition;
- root versus derived repository-path provenance;
- literal repository-relative join composition;
- exact `fs::{read,read_to_string,write}` qualification;
- Cargo alias -> `cargo run -p PACKAGE` resolution;
- default generator `src/main.rs` discovery;
- source -> generated-output relation assembly;
- exact `.gitattributes` generated-path discovery.

The first provider vocabulary remains deliberately narrow:

```text
project_root::get_project_root()
```

## Provenance controls

Canonical unit coverage requires:

```text
supported explicit root provider      -> accepted
arbitrary PathBuf parameter `root`    -> rejected
collection/string `.join(...)`        -> rejected
dynamic join suffix                   -> rejected
non-fs read/write lookalikes          -> rejected
literal `..` path escape              -> rejected
```

## Review-dogfood strengthening

After the first held-out Oxc matrix had already passed, the generated-provenance review dogfood found this counterexample:

```rust
let tasks = project_root::get_project_root()?.join("tasks");
let source = std::fs::read_to_string(tasks.join("src/rules.rs"))?;
```

The first implementation promoted `tasks` to repository root and lost the `tasks/` prefix.

The canonical model now distinguishes:

```text
RepositoryRoot
DerivedRepositoryPath("tasks")
```

so the example resolves to:

```text
tasks/src/rules.rs
tasks/generated/rules.rs
```

Final clean-head review dogfood:

```text
run 32221469968 — success
```

## Held-out Oxc product replay after the provenance repair

Exact target:

```text
```

First parent:

```text
568203e24f090f2cc3f945d611e605b864842bf0
```

The replay applied only the target's `crates/oxc_linter/src/rules.rs` patch to the parent and withheld both generated registries.

The canonical ownership view recovered:

```text
cargo lintgen -> run -p oxc_linter_codegen

rules.rs
  -> generated/rule_runner_impls.rs
  -> generated/rules_enum.rs
```

with exact `.gitattributes` generated evidence.

The public command:

```text
cargo-cultist diff --format json corpus/oxc
```

emitted exactly two `generated-companion-missing` findings, one for each missing registry. Each finding retained:

```text
derived
derived
observed
observed
unknown
```

with `99/99` semantic historical support and explicit `UNKNOWN` current intent / generated-byte effect.

## Exact historical negative control

The same product replay reconstructed source-only commit:

```text
5e113baf716b9f3781331b268b4142d23cac0541
docs(linter): add license notices for ported ESLint plugins (#22768)
```

Result:

```text
generated findings: 0
```

## Strong final corpus receipt

Executed post-dogfood semantic head:

```text
4796ed643e1ed8bfe4e0b8fe61df099214aee80d
```

Dedicated Oxc replay:

```text
run 32221343877
job 95972249274 — success
```

Artifact:

```text
9354031900
sha256:9143c1d2940fb5f056b814816253dcfc8569e362c93156d7963d870602534c8a
```

Full methodology is retained in:

```text
research/generated-provenance-restoration-replay.md
```

The temporary external replay workflow has been removed.

## Final source-only merge head

```text
3f841ecbea7ae2c3c7365f2c9efd20d3135ec23a
```

Generic CI:

```text
run 32221469937 — success
```

All substantive steps passed: rustfmt, Clippy `-D warnings`, full tests, repository/history/CI-test dogfood, and diff text + JSON.

Derived-root review dogfood:

```text
run 32221469968 — success
```

## Disposition

Restore the bounded generated-companion product finding. The canonical provenance implementation, original #80 acceptance matrix, stronger derived-root attack, held-out Oxc positive, exact docs-only negative, and final clean CI all pass.

Refs #80, #82, #70, #72, #68.


teamleaderleo (commented):
The proven-root semantics look aligned with #70/#80, but exact-head CI exposed an integration problem in the sharing boundary that I would fix rather than `allow(dead_code)` around it.

`src/generator_ownership.rs` is physically shared through two separate module inclusions. In the product compilation, `GeneratorSourceReport` / `analyze_generator_source` / research-only helpers are unreachable; in the example compilation, `GeneratorRelation` / `discover_generator_relations` / product-only helpers are unreachable. With `cargo clippy --all-targets -- -D warnings`, both halves fail as dead code.

I’d make the canonical adapter expose one analysis entry point/report that both consumers call, with product deriving relations from that report and the research view selecting its requested generator from the same report. Another clean option while the crate is unreleased is a tiny library module consumed by both binaries, but the important property is that we avoid two consumer-specific public APIs living in a separately compiled `#[path]` module.

This is useful dogfood evidence for #74 too: the during-code CI caught that “one source file” is not yet the same thing as “one coherent evidence adapter.”

teamleaderleo (commented):
There is a provenance hole beyond the current adversarial controls in `expression_uses_repository_root_provider`.

The visitor marks a local as a repository-root binding when its initializer contains the provider call anywhere in the expression tree. That means a derived path can be promoted back to root:

```rust
let tasks = project_root::get_project_root()?.join("tasks");
let source = std::fs::read_to_string(tasks.join("src/rules.rs"))?;
```

`tasks` currently gets inserted into `repository_roots` because its initializer contains `get_project_root()`. The later join can then be recorded as repository path `src/rules.rs`, silently dropping the real `tasks/` prefix. That manufactures the wrong ownership edge even though the provider itself is legitimate.

I’d change root recognition from “contains provider call” to “expression resolves to provider result through explicitly transparent wrappers.” For the Oxc witness that likely means recursively accepting the provider call itself plus `?` / parentheses and the known `map_err(...)` wrapper used around it, while rejecting path-transforming calls such as `.join(...)`. Add the derived-subdirectory case above as a negative regression.

This is the same core evidence rule #80 is trying to enforce: provenance includes transformations after the authoritative origin, not only whether that origin appears somewhere upstream.

teamleaderleo (commented):
One smaller drift remains inside the new canonical module: `discover_generator_relations()` uses `cargo_run_aliases()` / `package_from_cargo_run()` to resolve the package semantically, while `analyze_generator_source()` gets display aliases from `cargo_aliases_for_package()`, which currently filters with `alias.command.contains(package)`.

That can make the research view claim an alias names package `foo` when the command actually names `foo-extra` (or only mentions `foo` in some unrelated argument). Given this PR's goal, I'd derive the research report's package aliases through the same `package_from_cargo_run()` result used by product ownership, rather than retaining a second looser notion of “alias for this package.”

This does not look like the #80 product false-positive path itself, but it defeats part of the benefit of putting both consumers behind one evidence adapter.

teamleaderleo (commented):
Blocking review notes on this exact head:

1. **Derived path can still be promoted to repository root.** `expression_uses_repository_root_provider()` searches the whole initializer subtree. `let tasks = project_root::get_project_root()?.join("tasks");` therefore marks `tasks` as a root binding, and `tasks.join("src/rules.rs")` can be recorded as `src/rules.rs`, dropping the real `tasks/` prefix. Root provenance should survive only explicitly transparent wrappers around the provider result, not path-transforming calls such as `.join(...)`. Please add this as a negative regression.

2. **The shared boundary is currently papered over with `#![allow(dead_code)]`.** The prior exact-head Clippy failure showed product and research still use disjoint APIs from a separately compiled `#[path]` module. #80/#72 called for a canonical evidence implementation so semantics cannot drift; one coherent report/entry point used by both consumers would address the cause instead of suppressing the lint.

3. The smaller alias split also remains: research alias reporting uses `command.contains(package)` while product relations use parsed `cargo run -p PACKAGE` semantics. The canonical module should use the same package resolver for both.

The derived-root case is a product correctness blocker for #80; the other two are consolidation/evidence-consistency blockers.

teamleaderleo:
Executed adversarial proof for the derived-root review finding:

- workflow: `Generated provenance review dogfood`
- run: `32220950875`
- job: `95971168123`
- exact branch commit exercised: `3a40ff0667223c6f81040a70d0ce6da2aad0b5d4`

Fixture:

```rust
let tasks = project_root::get_project_root()?.join("tasks");
let source = std::fs::read_to_string(tasks.join("src/rules.rs"))?;
let target = tasks.join("generated/rules.rs");
std::fs::write(&target, source)?;
```

The canonical research view emitted:

```text
EXPLICIT PATH RELATIONS
  function generate
    reads  src/rules.rs
    writes generated/rules.rs
```

and the discriminator failed with:

```text
derived subdirectory was incorrectly promoted to repository root
```

So the current provider recognizer is demonstrably dropping the `tasks/` prefix. This should become a permanent negative unit regression in `generator_ownership.rs`; once embedded there, the temporary review workflow can be removed.

The useful evidence contract looks like: provider result may pass through explicitly transparent wrappers (`?`, parentheses, the known Oxc `map_err`) while path-transforming wrappers such as `.join(...)` stop root identity.

Issue #75, opened 2026-08-19
Research a repo-local decision record that feeds future agent context


teamleaderleo:
## Question

What is the smallest reviewable repo-local decision record that can close the agent memory loop from #74 without prematurely committing Cultist to a final `teach` UX or storage format?

This is a research slice of #10.

Desired loop:

```text
finding / investigation
-> explicit accepted rationale in version control
-> later target query
-> decision appears as applicable project memory
```

## First discriminator

Use Cultist itself as the corpus.

The repository already has an explicit project decision around historical co-change: raw companion frequency remains association evidence and must retain counterexamples/cohort receipts rather than becoming an automatic `must update` rule.

Represent that existing decision in a temporary research record, then prove a local resolver can surface it for `src/history.rs` while refusing it for unrelated targets.

## Research requirements

- deterministic/local/read-only resolver;
- one decision per file to reduce merge contention and preserve review history;
- stable record id;
- explicit scope separate from finding kind;
- human-readable rationale;
- exact source/authority references where available;
- optional expiry left out of v0 unless needed;
- invalid/unknown schema fails closed;
- broader/narrower scope matching is explicit;
- no implicit suppression: a decision is evidence first;
- model-generated text is not authoritative merely because it exists in a file on an unmerged branch.

## Candidate temporary record

```json
{
  "schema_version": 1,
  "id": "history-cochange-remains-association-v1",
  "kind": "historical-companion-policy",
  "scope": {"path_prefix": "src/history.rs"},
  "reason": "Raw co-change support is association evidence; frequency alone does not establish a required companion update.",
  "authority": [
    {"kind": "issue", "reference": "#34"},
    {"kind": "pull_request", "reference": "#39"}
  ]
}
```

The exact field names/file location are research hypotheses only.

## Success criteria

- resolver returns the record for the intended target;
- unrelated target does not inherit it;
- output preserves record id, reason, scope, and authority refs;
- malformed/unsupported records are reported explicitly rather than silently ignored;
- a future #62 `brief` implementation could consume the same resolved object without a second memory store;
- record remains evidence and does not automatically suppress a finding.

Refs #10, #18, #62, #74.

teamleaderleo:
Accepted-ref provenance research negative receipt from stale PR #95:

- branch `research/decision-memory-accepted-ref-provenance@7fbfd6ca...` added only a temporary research workflow + current-only fixture;
- ordinary fmt/clippy/example tests passed;
[...]

PR #95, opened 2026-08-19
research: classify decision records against an accepted Git ref

teamleaderleo:
## Question

Can repo-local decision memory distinguish exact record content already present on an accepted Git reference from record content that exists only in the current branch, while keeping unavailable-reference uncertainty explicit?

This attacks the authority-provenance promotion gap after second-kind, longitudinal, and rename-survival research.

## Semantics

Each resolved decision gains:

```text
acceptance.status:
  accepted_ref
  current_only
  unknown

acceptance.reference:
  <configured Git ref>
```

Default accepted reference:

```text
refs/remotes/origin/main
```

Research/replay can override it with:

```text
CARGO_CULTIST_DECISION_ACCEPTED_REF
```

These statuses make a Git-content claim only:

- `accepted_ref`: the configured ref exists and contains identical bytes at this record path;
- `current_only`: the ref exists, but identical current bytes are absent there;
- `unknown`: the configured ref itself is unavailable locally.

They do not claim a human review ceremony occurred.

## Live branch matrix

This PR deliberately adds a temporary fixture decision scoped to `src/main.rs` that does not exist on `origin/main`.

The workflow explicitly fetches `main` and requires:

```text
history-cochange-remains-association-v1
  -> accepted_ref

generated-companion-canonical-provenance-v1
  -> accepted_ref

current-only-authority-fixture-v1
  -> current_only
```

Then it overrides the accepted ref to an intentionally missing remote ref and requires the existing history decision to become:

```text
unknown
```

## Boundary

`accepted_ref` means exact record bytes are present on the configured Git reference. It does not establish who reviewed the record, whether approvals were required, or whether the accepted ref itself is governed by branch protection.

The temporary current-only fixture will be removed after a successful receipt. Agent-packet schema will be bumped separately once the standalone acceptance matrix succeeds.

### DIFF
--- a/.github/workflows/decision-memory-accepted-ref-research.yml
+++ b/.github/workflows/decision-memory-accepted-ref-research.yml
@@ -1,0 +1,121 @@
+name: Decision memory accepted-ref provenance research
+
+on:
+  pull_request:
+    paths:
+      - examples/decision_memory.rs
+      - research/decision-memory/current-only-authority-fixture.json
+      - .github/workflows/decision-memory-accepted-ref-research.yml
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  accepted-ref:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout cargo-cultist
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Fetch accepted main reference
+        run: |
+          git fetch origin main:refs/remotes/origin/main
+          git rev-parse --verify refs/remotes/origin/main^{commit}
+
+      - name: Install stable Rust
+        run: |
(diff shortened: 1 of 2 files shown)


teamleaderleo:
Retiring this carrier as a failed/stale research experiment, not as rejection of the accepted-ref provenance question.

Dedicated run `32224526619` showed the workflow contract had advanced ahead of the implementation: it asserted `accepted_reference` / `acceptance` fields that the resolver did not emit, failing with `KeyError: 'accepted_reference'`. Current main still has no accepted-ref/current-only decision-memory semantics.

The negative receipt is preserved on #75. A future replay should restart from current main with an explicit minimal acceptance claim rather than carrying this temporary workflow/fixture forward.

Closing also keeps the live active-work inventory from treating this failed temporary lane as ongoing work.

Issue #101, opened 2026-08-19
Proposal: explicit coordination edges via offline active-change inventories


teamleaderleo:
## Thesis

Direct path overlap is only the baseline for `cultist preflight`. The next high-precision layer should consume **project-authored coordination evidence** before adding statistical history heuristics.

Real agent-heavy repository history already contains explicit sequencing and dependency language such as:

- `depends on #N`;
- `blocks #N`;
- `hold merge behind #N`;
- `do not merge while #N ...`;
- `supersedes #N`;
- `head movement expires evidence`;
- exact base/head evidence coordinates.

These relations can require coordination even when two changes have zero shared paths.

The strongest public discriminator from #99 is `teamleaderleo/preflight` #748 vs #703: #748 explicitly says not to merge while #703 is using current-main package evidence, yet the two PRs touch disjoint path sets. Plain Git sees independence; project-authored evidence says merge order matters.

## Reuse the landed active-work contract

#98 already proved the desired split architecture in live dogfood:

```text
remote adapter
  -> schema-versioned active-work inventory

local Rust analyzer
  -> validate + normalize
  -> evidence analysis
```

Do **not** introduce a second active-change inventory format.

Promote the landed schema-v1 snapshot used by `examples/active_work_heads_up.rs` and `.github/workflows/active-work-heads-up-research.yml`:

```json
{
  "schema_version": 1,
  "source": "github_pull_requests",
  "observed_at": "2026-08-19T00:01:00Z",
  "current": {
    "id": "#748",
    "kind": "pull_request",
    "title": "...",
    "url": "...",
    "head_ref": "...",
    "head_sha": "...",
    "updated_at": "...",
    "draft": false,
    "changed_paths": ["path/a"]
  },
  "active_work": [],
  "coordination_edges": [
    {
      "kind": "hold_merge_while",
      "from": "#748",
      "to": "#703",
      "source": "github:pull/748"
    }
  ]
}
```

`coordination_edges` is an optional backward-compatible enrichment. Existing #98 snapshots with no such field remain valid input.

## Product slice

Add:

```text
cargo cultist preflight --inventory FILE [--format text|json] [PATH]
```

`--against REV` remains the local arbitrary-ref mode. The two modes are mutually exclusive.

Inventory mode should:

1. read a bounded local JSON file;
2. strictly validate the landed schema-v1 work identity/freshness/path fields plus optional coordination edges;
3. exclude the current work when it reappears in `active_work`, as the landed adapter does;
[...]

Issue #96, opened 2026-08-19
Proposal: preflight collision analysis for concurrent human/agent changes


teamleaderleo:
## Thesis

Cultist already asks a repository what precedent, history, generated relationships, CI selectors, and declared policies say about one change. The next natural step is to compare **multiple concurrent changes before Git reports a merge conflict**.

The target is a preflight check that answers:

> What else is this change likely to collide with, even when Git can merge the text cleanly?

This is especially useful for coding agents. Agents can produce many branches in parallel while lacking the ambient team awareness humans carry across time: who is rewriting a subsystem, which files are generated together, which paths historically move together, which policy another branch is changing, and which local convention is already under active revision.

## Collision species

Keep these distinct in the evidence model.

### 1. Direct path collision

Two changes modify the same repository path.

This is deterministic and should be reported as `PROVEN`.

### 2. Generated / ownership collision

One change modifies a source while another modifies or omits a generated companion, generator, or declared output.

Use existing generated-relationship evidence only after its product provenance gates are repaired. Never infer ownership from co-change alone.

### 3. Historical companion collision

Two changes modify different paths that repeatedly moved together in comparable history.

This remains `OBSERVED` association evidence. Preserve direction, opportunities, support, counterexamples, and cohort limitations. Do not turn association into a required-update rule.

### 4. Policy collision

One change modifies behavior while another modifies a repository-declared policy or oracle governing that behavior: lock ranks, CI selectors, generator declarations, package metadata, etc.

Prefer explicit repository policy over historical popularity.

### 5. Convention / precedent collision

Parallel changes introduce competing local conventions or both alter an area where file-local and repository-wide precedent already disagree.

Surface the tension and let humans/agents choose deliberately.

### 6. Intent collision

Two branches appear to implement competing approaches to the same repository responsibility.

This begins as `UNKNOWN` / `INFERRED` unless explicit issue, PR, decision-record, or code evidence supports a stronger claim.

## Product idea

A local command along these lines:

```text
cargo cultist preflight --against REF [PATH]
```

Initial semantics:

- current side = working tree changes from the merge base of `REF` and `HEAD`, including staged/unstaged changes;
- other side = commits reachable from `REF` since that same merge base;
- compare repository-relative changed paths;
- emit direct overlap as `PROVEN`;
- enrich different-path pairs with independent repository evidence where a supported analyzer can establish a relationship;
- text and JSON render from the same finding model;
- remain deterministic, local, and read-only.

[...]

PR #107, opened 2026-08-19
research: retain divergent branch heads-up adapter

teamleaderleo:
Research follow-up under #96 for live work that exists **before a PR**.

## What remains in this PR

The temporary replay workflow is removed. The durable artifacts are:

- `scripts/divergent_branch_inventory.py` — bounded PR + divergent-branch inventory adapter;
- `research/divergent-branch-heads-up.md` — executed receipts, lifecycle failure, quietness matrix, and product boundary.

The adapter reuses the landed #98/#100 work-item schema and the product #103 `preflight --inventory` contract. It does not invent a second product format.

## Branch admission

A remote branch is not surfaced merely because it exists or has a relevant-looking name.

The research adapter:

1. starts from the open-PR inventory;
2. excludes branch heads already represented by open PRs;
3. discovers refs not merged by Git ancestry into `origin/main`;
4. retires an exact current branch `name + head SHA` when that same exact head appears among the recently observed closed/merged PR heads;
5. orders remaining candidates by exact commit timestamp;
6. returns at most 20 and records omissions;
7. measures exact paths from each branch's merge base;
8. records that branch-name similarity was not used.

The exact-head retirement rule matters because squash-merged PR branches are not ancestors of `main`.

## Dogfood failure that improved the model

The first run (`32232628351`, job `96005566152`) failed because the original control assumed `feature/preflight-active-inventory` would be excluded by Git ancestry.

PR #103 had been squash-merged. Its source branch therefore still looked `--no-merged` even though the work had been accepted.

The repair combines Git divergence with provider lifecycle:

```text
Git says histories have not merged
+ GitHub says this exact branch head already belonged to a closed/merged PR
-> retire that exact head

branch advances to a new SHA later
-> eligible again
```

## Final executed receipt

Run `32233054642`, job `96006879211` succeeded.

Artifact:

```text
id:     9357939745
sha256: 45860982f7ed855b5f992b894a9360294cf16110ce7283f603ebb034e67e16a9
```

Observed adapter receipt:

```text
open PR heads excluded:                3
recent closed/merged PR heads seen:   48
closed-PR head window truncated:      false
exact closed-PR branch heads retired: 23
unmerged non-PR branch candidates:     7
branch candidates returned:            7
branch candidates omitted:             0
max branch candidates:                20
branch-name similarity used:          false
```

### Natural quietness

The #107 carrier stayed quiet with the combined PR + branch inventory.

The same inventory, after removing research-only adapter receipts, was admitted by product:

```text
cargo cultist preflight --inventory FILE --format json .
```

and produced zero findings for the disjoint carrier.

### Pre-edit focus positive

Before the carrier touched the adapter path, focus mode surfaced the still-unpublished work:

```text
branch:experiment/divergent-branch-heads-up
d99c27e5
2026-08-19T16:28:01+08:00
focus overlap: scripts/divergent_branch_inventory.py
```

### Noise matrix

Replaying the same snapshot with every open PR as current produced:

```text
#107 heads-up count: 0
#102 heads-up count: 0
#95  heads-up count: 0
```

Seven branch candidates existed, but none interrupted the three open PRs in that snapshot.

## Why this is not auto-enabled yet

The observed repo was fully covered by the recent-100 closed/merged-PR window: only 48 heads were returned, so the lifecycle receipt was not truncated.

In a larger repository that bound can truncate. Until we dogfood a targeted branch-to-PR lifecycle lookup or another corpus beyond the window, this stays an explicit research adapter rather than silently broadening the always-on PR advisory.

## Product boundary

Product `preflight --inventory FILE [PATH]` treats `[PATH]` as **scope over observed current changes**. The research focus path is **caller-supplied pre-edit intent**. We deliberately do not falsify `current.changed_paths` to make pre-edit focus work.

A future product-facing focus/intent interface should keep those claims separate.

Refs #96, #98, #100, #101, #103, #62, #74.

### DIFF
--- a/research/divergent-branch-heads-up.md
+++ b/research/divergent-branch-heads-up.md
@@ -1,0 +1,239 @@
+# Divergent branch heads-up research
+
+Date: 2026-08-19
+
+## Question
+
+Can Cultist treat **unpublished divergent remote branches** as a second live-work source without confusing old/squash-merged branches with active work or increasing the normal interruption rate?
+
+This extends the PR-only active-work experiment from #98/#100 under the broader concurrent-change proposal in #96.
+
+The branch layer stays advisory and reuses the same `WorkItem` inventory vocabulary:
+
+```text
+open PR metadata
++ bounded divergent remote branches
+-> active-work inventory
+-> exact path overlap / pre-edit focus
+```
+
+Branch-name similarity is never used as evidence.
+
+## Adapter candidate
+
+The research adapter lives in `scripts/divergent_branch_inventory.py`.
+
+It starts from the landed GitHub PR inventory and adds remote branches only when they satisfy explicit lifecycle/evidence gates.
+
+### PR heads win over duplicate branch identities
+
+If an open PR already represents a branch head, the branch is excluded. The richer PR item carries provider URL, draft state, and PR freshness, so a second `branch:*` work item would be duplicate evidence.
+
+### Git divergence is necessary but not sufficient
+
+Candidate refs come from:
+
+```text
+git branch -r --no-merged origin/main
+```
+
+and their changed paths are measured from the exact merge base with `origin/main`.
+
+This catches real unpublished branch work, but the first dogfood exposed a major Git-only lifecycle trap: **squash-merged PR heads remain topologically unmerged**.
+
+PR #103 was squash-merged to `main`; its source branch `feature/preflight-active-inventory` therefore still appeared under `git branch --no-merged` even though the work had already been integrated.
+
+The first research run failed precisely because the original negative control assumed Git ancestry would retire that branch.
+
+### Provider lifecycle repairs squash-merge ambiguity
+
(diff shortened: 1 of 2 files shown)


teamleaderleo:
Research-path fidelity refinement: `changed_paths()` currently uses line-delimited `git diff --name-only` and `git_lines()` strips every returned line. That means the adapter's claimed exact paths are not byte-faithful for unusual-but-valid Git filenames: leading/trailing whitespace can be altered, and tab/newline/control-containing paths are C-quoted by normal Git output instead of yielding the repository path. Since this adapter is explicitly measuring exact path overlap, use NUL-delimited Git output (`git diff -z --name-only ...`) with a NUL parser that does not `.strip()` path bytes/text. The same helper can stay for metadata commands where trimming is deliberate. A focused fixture with a leading/trailing-space filename and a tab/newline filename would make the evidence contract explicit.

Issue #119, opened 2026-08-19
Prototype terse agent output with expandable evidence references


teamleaderleo:
## Goal

Prototype the first operational projection from #115:

> store richly, transmit minimally, expand on demand.

Do not invent the complete Cultist IR in this issue. Reuse the existing typed finding/report model and prove that an agent-facing renderer can be dramatically smaller than prose while retaining a deterministic path back to full evidence.

## Product question

Can a routine Cultist result communicate only what changes the caller's next action while preserving provenance, uncertainty, and debugging detail behind stable references?

## First slice

Add a research/experimental rendering mode over one or two existing commands, ideally `diff` and/or `preflight`:

```text
--format terse
```

The exact flag is provisional.

Candidate output classes:

```text
OK <report-ref>
```

```text
FIND <finding-ref> <kind>
```

```text
HOLD <unknown/ref>
NEXT <required discriminator/action>
```

```text
PASS <check-ref>
LIMIT <evidence-boundary>
```

The renderer must be derived from the same report/finding object used by text/JSON. It must not create a second state model.

## Stable expansion path

Every terse reference must be resolvable within the emitted object/context.

Prototype one deterministic expansion interface, for example:

```text
cultist inspect <ref> --format json
cultist inspect <ref> --format text
```

or an equivalent report-local reference mechanism.

If durable cross-run IDs are premature, begin with a report fingerprint + local typed ID and make the lifetime explicit.

## Silence / interruption rule

For the experimental renderer, omit routine narration unless the event does at least one of:

- changes the caller's next action;
- introduces or resolves material uncertainty;
- invalidates an assumption/evidence coordinate;
- requires escalation or another discriminator;
- exposes a finding worth acting on;
- creates a durable evidence reference the caller requested.

Do not emit congratulations, prose restatements of exit codes, repeated scope headers, or explanations of already-referenced deterministic facts.

## Evidence boundaries remain visible

Compression must not turn:

```text
target test passed on Linux only
```

into an unqualified `PASS` if the platform limit affects the decision.

Likewise, `UNKNOWN`, stale evidence, omitted evidence, and unresolved semantic independence must remain explicit when material.

## Suggested experiment

Use a small corpus of real Cultist/Stensibly cases and compare:

1. current text output;
2. current JSON output;
3. terse output + on-demand expansion.

Measure:

- bytes/tokens emitted initially;
- bytes/tokens after any requested expansions;
- whether fresh agents choose the same correct next action;
[...]

Issue #115, opened 2026-08-19
Research a compact Cultist IR and context-relative grammar


teamleaderleo:
## Thesis

Cultist should research a **small domain grammar / intermediate representation (IR)** for repository evidence, work envelopes, review state, decisions, uncertainty, and temporal transitions.

The goal is not to invent a general-purpose language. It is to represent a bounded set of repository concepts more compactly and precisely than repeated natural-language narration.

The same underlying meaning should support multiple projections:

```text
                  Cultist IR
                      |
        +-------------+-------------+
        |             |             |
        v             v             v
   compact agent     JSON/debug     human view
     protocol          view        concise prose
```

Human-readable text is one renderer. It should not be the canonical storage format merely because current code-review and issue interfaces are prose-shaped.

Refs #22 #62 #67 #74 #106 #109.

## Representation principle

For the bounded concepts Cultist actually owns, two representations can be losslessly translated when they preserve the same distinctions and references. Do not assume every arbitrary representation is literally bijective; define the semantic domain first, then test round-trip preservation inside that domain.

Candidate semantic basis:

```text
IDENTITY
SCOPE
CLAIM
EVIDENCE
RELATION
COUNTEREXAMPLE
UNKNOWN
ASSUMPTION
VALIDITY
DECISION
DISPOSITION
INVALIDATION
SUPERSESSION
REOPEN
OMISSION
FRONTIER
TRANSITION
```

This list is a research hypothesis, not a frozen vocabulary.

The useful question is:

> What is the smallest stable set of typed distinctions from which Cultist's current evidence, JEI, review, and memory objects can be reconstructed without repeatedly restating their meaning in prose?

## Ambient context should do work

Agents usually enter a repository with substantial implicit context already available:

- repository identity;
- current or requested revision;
- task/work identity;
- Cultist schema/version;
- current envelope/review object;
- known project decisions/evidence graph.

Do not retransmit information that is already unambiguous from that context.

A message may begin with a small explicit context header:

```text
schema: cultist-ir/v1
repo: teamleaderleo/cultist
rev: <exact head>
task: <task identity>
base: <optional prior envelope fingerprint>
```

After that, references can be relative when the decoder can resolve them deterministically.

For example, instead of:

```text
The current change at commit abc modifies src/foo.rs.
Reviewed decision #123 applies to src/foo.rs.
Exact-head target execution has not occurred.
Hold review until that execution exists.
```

an IR might represent the same bounded state conceptually as:

```text
S P17
[...]

PR #121, opened 2026-08-19
research: prototype terse report rendering

teamleaderleo:
## Summary

First code experiment for #119 / #115, still intentionally below public CLI/protocol commitment.

- add `render_terse_analysis_report(&AnalysisReport)` alongside the existing human text renderer;
- keep the same canonical `AnalysisReport` object as the only semantic source;
- render a no-finding/non-UNKNOWN report as literal `NO_FINDINGS` rather than inventing an `OK`/approval disposition;
- surface findings with deterministic report-local `F1`, `F2`, ... references;
- surface claim provenance with compact `P` / `D` / `O` / `I` / `?` tokens;
- retain finding location, claim messages, material `UNKNOWN`, and the current question;
- omit supporting evidence prose from the terse projection so it remains available through richer renderers instead of being retransmitted every time;
- add controls for clean/no-finding output, claim-only no-finding output, report-level UNKNOWN, deterministic local refs, provenance tokens, and evidence-prose omission.

## Projection boundary

The renderer may omit information, but it must not strengthen the canonical report while doing so.

`AnalysisReport` currently has no typed pass/approved/ok disposition. Therefore:

```text
findings == []
```

can justify `NO_FINDINGS`; it does not by itself justify `OK`.

An adversarial control now supplies zero findings plus a nontrivial INFERRED report-level claim and requires the terse projection to remain `NO_FINDINGS`, never an approval-like token.

## Important boundary

This PR deliberately does **not** add `--format terse` yet and does not claim durable IDs.

`F1`, `F1.C2`, etc. are report-local only. The current machine schema has no stable claim/finding identities, so cross-run expansion should not be invented until #115 earns an identity/fingerprint contract.

The renderer/helper are explicitly marked research-only so Clippy does not force premature CLI exposure merely to make the experiment reachable in non-test builds.

## Why this composes rather than competes

This is a projection over the same `AnalysisReport` used by text/JSON. It does not create a second evidence model, selection policy, or authority layer:

- C1 asks how the full report can be structurally carried;
- JEI asks what evidence should be selected;
- review envelopes ask where attention should go;
- terse rendering tests how little of an already-selected report can be transmitted without strengthening meaning.

Refs #115 #119 #116 #120 #122.

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -18,6 +18,62 @@ pub fn render_analysis_report(report: &AnalysisReport) -> String {
     output
 }
 
+/// Experimental agent-oriented projection over the same canonical AnalysisReport.
+///
+/// References are deliberately report-local in v0: F1 is the first finding,
+/// F1.C2 is its second claim, and C1 is the first report-level claim. They are
+/// not durable IDs and must not be persisted as cross-run identity.
+///
+/// This function is research-only until a public terse format and expansion
+/// contract earn promotion.
+#[allow(dead_code)]
+pub fn render_terse_analysis_report(report: &AnalysisReport) -> String {
+    let mut output = String::new();
+    let mut emitted = false;
+
+    for (index, claim) in report.claims.iter().enumerate() {
+        if claim.kind == ClaimKind::Unknown {
+            writeln!(output, "U C{} {}", index + 1, claim.message).unwrap();
+            emitted = true;
+        }
+    }
+
+    for (finding_index, finding) in report.findings.iter().enumerate() {
+        let finding_ref = format!("F{}", finding_index + 1);
+        write!(output, "{finding_ref} {}", finding.kind).unwrap();
+        if let Some(location) = &finding.location {
+            write!(output, " @{}", format_location(location)).unwrap();
+        }
+        output.push('\n');
+
+        for (claim_index, claim) in finding.claims.iter().enumerate() {
+            writeln!(
+                output,
+                "  C{} {} {}",
+                claim_index + 1,
+                claim_kind_token(claim.kind),
+                claim.message
+            )
+            .unwrap();
+        }
+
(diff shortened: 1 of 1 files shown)


teamleaderleo:
Review found one mechanical blocker and one semantic boundary worth fixing before this projection earns promotion.

### Mechanical

CI run `32238256463` fails Clippy because the experimental renderer is intentionally not wired into the CLI yet, so `render_terse_analysis_report` / `claim_kind_token` are dead in the non-test binary build. That is easy to contain for a research-only function (`#[allow(dead_code)]` / isolated example/module / another non-public call site); it should not force a public `--format terse` just to satisfy Clippy.

### Semantic: `OK` is stronger than `AnalysisReport`

The existing report model has findings, claims, provenance, evidence, and questions. It does **not** have a typed `ok/passed/approved` disposition.

So:

```text
findings == []
+ no report-level UNKNOWN
-> `OK`
```

adds meaning the source object did not prove. A report may still contain nontrivial PROVEN/DERIVED/OBSERVED/INFERRED top-level claims, and absence of a finding is not the same claim as “the change is okay.” This is the same projection discipline we are now enforcing elsewhere: a compact view may omit data, but it should not strengthen semantics while doing so.

For this first slice I would prefer an exact projection token such as:

```text
NO_FINDINGS
```

(or similarly literal wording) until a later JEI/review/disposition contract actually earns `OK` / `PASS` semantics.

Please add a negative control with **zero findings + a nontrivial top-level non-UNKNOWN claim** and require that the terse result does not imply a stronger disposition than the canonical report contains. Report-level UNKNOWN should of course remain explicit as already intended.

This does not make the terse concept competing with C1/JEI/review. It is a useful projection over the same report; the constraint is simply `projection <= source semantics`.

Refs #115 #119 #121 #116 #120.

Issue #123, opened 2026-08-19
Research shared evidence applicability and validity semantics


teamleaderleo:
## Thesis

Cultist now has multiple features that need to answer the same question:

> Does this evidence still apply to this exact repository/revision/work/scope?

Active work, JEI, review envelopes, decision memory, preflight, compact IR, and future deltas should reuse one typed applicability model instead of each inventing its own freshness/acceptance flags.

This issue owns **coordinate applicability / invalidation semantics**. It does not own claim truth strength, source authority, action disposition, JEI selection, or transport encoding.

Refs #10 #62 #67 #74 #75 #96 #106 #109 #115 #119 #120 #122.

## Keep four axes orthogonal

```text
EPISTEMIC
  proven / derived / observed / inferred / unknown

APPLICABILITY
  applies / invalid / unknown

AUTHORITY
  source identity + accepted/reviewed trust event where one exists

DISPOSITION
  ok / hold / next / reopen / escalate / ...
```

Examples:

- an OBSERVED provider path set can be inapplicable after the provider head moves;
- an authoritative decision can be out of scope for a different path;
- an explicit PR-body clause can be genuine source evidence while its broader applicability to the current diff remains UNKNOWN;
- a passing test can be current and PROVEN without granting merge authority.

Do not collapse these into one confidence/status field.

## First deterministic slice: exact coordinate applicability

Before wall-clock freshness or policy expiry, model exact constraints that can be evaluated without heuristics.

Candidate context dimensions:

```text
repository
revision
work
scope/path
```

A piece of evidence carries the dimensions it requires. A current query supplies whatever dimensions it actually knows.

Evaluation:

```text
all required dimensions present + match
  -> APPLIES

required dimension present + mismatches
  -> INVALID

required dimension missing from current context
  -> UNKNOWN
```

Every result should name the matching/mismatching/missing dimensions. Missing current context must never be silently treated as a match.

## Scope semantics

Start deliberately small:

- repository: exact string identity;
- revision: exact object/revision identity;
- work: exact provider/project work identity;
- path scope: canonical repository-relative exact path or path-prefix containment.

Reject absolute/traversing/noncanonical paths rather than guessing.

Do not add semantic-symbol/package/time-range matching until a real discriminator requires it.

### A. Provider head movement

Evidence observed for work `#748` at head `a2e14c...`:

- same work + same head -> applies;
- same work + new head -> invalid for exact-head claims;
- current head unavailable -> unknown.

[...]

Issue #109, opened 2026-08-19
Research review intelligence and JEI review envelopes


teamleaderleo:
## Thesis

Cultist should treat code review as an **evidence-allocation problem**.

The scarce resource is reviewer attention. Before useful judgment can happen, a reviewer has to understand the change, find relevant project context, decide which parts deserve scrutiny, separate mechanical feedback from ambiguity, and notice when specialized knowledge or deeper investigation is needed.

Cultist already has many of the required primitives: JEI (#106), scoped precedent/counterexamples, history, active-work evidence, decision memory, provenance, and demand-driven analysis.

Research question:

> Can Cultist compile a small, current, provenance-bearing review envelope that helps a reviewer understand what changed, where attention is most valuable, what evidence should affect judgment, and which review outcomes should survive as future project memory?

## Candidate review envelope

```text
identity
  exact change coordinate + envelope fingerprint

change decomposition
  related review clusters
  independent-looking regions with explicit UNKNOWN where independence is unproven

attention
  selected high-value review regions
  evidence explaining why they were selected
  omission receipts for mechanical / lower-priority regions

risk / uncertainty
  deterministic risk signals
  active-work collisions
  guidance/precedent contradictions
  missing validation / evidence
  material UNKNOWNs

project context
  accepted decisions
  explicit guidance
  scoped precedent + counterexamples
  relevant historical failures / prior review rationale

mechanical evidence
  deterministic findings / tests / static checks
  routine autofix candidates

review frontier
  questions still requiring judgment
  missing discriminator
  specialized expertise needed
  stop / escalate conditions

post-review memory
  concerns that changed the patch
  concerns rejected by counterexample
  accepted rationale
  repeated review consensus eligible for later promotion
```

A review envelope grants no merge or external-effect authority.

## Lane 1: change decomposition

Research deterministic decomposition before review using the cheapest justified relations:

- path/module/package;
- generated ownership;
- historical companions;
- changed items/symbols where available;
- source/test or schema/consumer relationships;
- explicit issue/PR intent.

Do not infer semantic independence from path disjointness alone.

Goal: help a reviewer see conceptual slices and accidental bundled work without reading the entire diff as one undifferentiated object.

## Lane 2: review-attention selection

JEI for review should answer:

> Which parts of this change deserve scarce review attention, and why?

Candidate signals include:

- first exception in a precedent cohort;
- explicit guidance conflicting with the change;
- historical failure/repair lineage;
- changed trust/authority boundary;
[...]

Issue #106, opened 2026-08-19
Research JEI work envelopes and a live repository research frontier


teamleaderleo:
## Thesis

Cultist should optimize for **just enough information (JEI)**: the smallest current evidence set that lets a worker act well, understand the important uncertainty, and know when to stop or escalate.

This is different from maintaining an exhaustive project snapshot or writing increasingly long agent instructions.

A static snapshot can be comprehensive and still become stale or drown the worker in irrelevant detail. Cultist already has the beginnings of something more useful: it can derive bounded evidence from the **current repository state, history, active work, accepted decisions, explicit guidance, counterexamples, and evidence freshness**.

The research question is:

> Can Cultist compile a small, current, provenance-bearing work envelope that describes what a task requires the worker to know — and separately expose the repository's live research frontier where important knowledge is still missing?

## Why this matters for different workers

Do not encode model brands or a permanent `small model` / `big model` taxonomy.

Instead describe the **epistemic demands of the work**:

- how bounded the target is;
- which facts are already established;
- which repository scopes are relevant;
- whether counterexamples exist;
- whether current evidence is fresh enough;
- whether a deterministic completion check exists;
- whether semantic interpretation is required;
- which unknowns are material to the requested action;
- which adjacent regions are supported exploration versus unsupported wandering;
- which conditions should stop the current lane and trigger deeper investigation.

A worker or orchestrator can then choose how much capability to spend on the task.

The same JEI packet should help a strong worker too: avoiding unnecessary repository archaeology is useful regardless of model capability.

## Candidate work-envelope contract

A first research object could look conceptually like:

```text
identity
  repository / revision / working-tree identity
  target / task identity
  envelope schema + fingerprint

objective
  bounded requested outcome
  explicit non-goals when repository evidence supports them

scope
  primary paths/items
  relevant repository/package/module scopes
  supported expansion regions

evidence
  PROVEN / DERIVED / OBSERVED claims
  accepted decisions / explicit guidance
  current relevant findings
  historical companions + counterexamples
  active-work / collision evidence

uncertainty
  material INFERRED claims
  UNKNOWN facts
  stale or unavailable evidence
  omitted/truncated evidence receipts

completion
  deterministic checks or observable success conditions
  evidence expected from the worker

stop / escalate
  conditions that invalidate the current bounded lane
  missing semantic capability
  guidance/precedent contradiction
  stale evidence coordinate
  scope expansion beyond supported region

research frontier
[...]

Issue #67, opened 2026-08-19
Make agent context packet omission and byte budgets explicit


teamleaderleo:
## Context

Draft PR #63 successfully replayed a local-only agent context packet against SmolRunner. The research result is useful: target-local history recovered earned repairs and companion surfaces, while remote rationale remained explicit `UNKNOWN`.

Before a public `cargo cultist brief` interface, the packet budget contract needs to become exact.

## Current gap 1: relation examples are silently truncated

The research example caps both examples and counterexamples per historical relation with `max_examples_per_relation`, but the packet does not currently report how many examples/counterexamples were omitted for each relation.

A consumer therefore cannot distinguish:

```text
2 examples exist total
```

from:

```text
40 examples exist; 38 were omitted
```

For an agent-facing packet, omission must never look like negative evidence.

### Desired behavior

Each relation should expose either exact totals or omission counts, for example:

```json
{
  "examples": [...],
  "examples_omitted": 12,
  "counterexamples": [...],
  "counterexamples_omitted": 3
}
```

or an equivalent deterministic truncation receipt.

## Current gap 2: the prototype is count-bounded, not byte-bounded

The research design proposes a byte budget and an evaluation metric around useful evidence per packet byte. The current `PacketBudget` bounds history commits, companion paths, examples, and broad-commit size, but it does not cap serialized packet bytes.

That is a reasonable v0 research boundary. Before public promotion:

- add a deterministic serialized-size budget;
- make every eviction observable;
- preserve high-priority evidence classes and counterevidence/UNKNOWN according to explicit ordering;
- record final serialized byte size;
- add adversarial fixtures with long paths/commit subjects/guidance metadata so count limits cannot silently exceed the intended envelope.

## Evaluation consequence

Once exact omission and byte counts exist, the proposed metric becomes measurable:

```text
useful evidence consulted per packet byte / developer interruption
```

Until then, describe the v0 packet precisely as **count-bounded research output**.

Refs #63, #62, #6, #15, #22.

teamleaderleo:
PR #73 merged the first half of this issue.

Now shipped in the research packet (schema v2):

- exact `examples_omitted` and `counterexamples_omitted` per historical relation;
- packet-level receipt when returned relations contain omitted examples/counterexamples;
- exact pretty-printed JSON `serialized_bytes`;
- executable acceptance proving omission algebra and byte measurement;
- explicit `UNKNOWN` that no hard byte budget / eviction policy exists yet.

[...]

PR #124, opened 2026-08-19
research: add shared evidence applicability semantics

teamleaderleo:
Starts #123 and composes with #75/#106/#109/#115.

## What

Add an isolated exact-coordinate applicability evaluator that answers whether a piece of evidence still applies to the current repository/revision/work/path context.

The first model keeps four concerns deliberately separate:

```text
epistemic strength   -> existing P/D/O/I/U claim provenance
applicability        -> this PR
source authority     -> separate evidence/acceptance semantics
disposition/action   -> separate review/orchestrator semantics
```

## Exact applicability dimensions

A requirement may constrain:

- exact repository identity;
- exact revision/evidence coordinate;
- exact work identity;
- canonical repository-relative path scope (`exact` or explicit `prefix`).

Evaluation is deterministic:

```text
all required dimensions present and match -> applies
known mismatch                            -> invalid
required context missing                  -> unknown
```

A known mismatch wins over a simultaneous missing dimension because applicability is already disproved.

Every evaluated dimension keeps a matched/mismatched/missing receipt with required and observed values.

## Controls

The standard `cargo test` matrix now exercises:

- public Preflight-style work/head coordinate applies;
- provider head movement invalidates exact-head evidence;
- unavailable current head stays UNKNOWN;
- mismatch takes precedence over missing context;
- exact and prefix path scope remain distinct;
- decision scope refuses unrelated targets;
- unresolved future delta/base coordinate stays UNKNOWN;
- dimension evaluation order is stable;
- absolute/traversing/backslash/noncanonical paths reject;
- unknown schema and unknown JSON fields fail closed;
- query and evaluation objects round-trip through serde.

## Research example

```text
cargo run --example evidence_applicability < query.json
```

The example admits at most 1 MiB and prints the typed evaluation as JSON.

## Boundary

- no wall-clock freshness/expiry yet;
- no network/provider calls;
- no source authority field;
- no `OK`/`HOLD`/merge disposition;
- no suppression or mutation;
- no primary CLI change;
- no changes to active #121 files.

This is meant to become one reusable primitive for active-work receipts, decision scope, JEI selection, review validity, and fail-closed envelope deltas only after the exact-coordinate semantics survive dogfood.

Refs #10 #62 #67 #74 #75 #96 #106 #109 #115 #119 #120 #122 #123.

### DIFF
--- a/examples/evidence_applicability.rs
+++ b/examples/evidence_applicability.rs
@@ -1,0 +1,37 @@
+#[allow(dead_code)]
+#[path = "../src/applicability.rs"]
+mod applicability;
+
+use std::error::Error;
+use std::io::{self, Read};
+
+use applicability::{
+    ApplicabilityQuery, MAX_APPLICABILITY_QUERY_BYTES, evaluate_query,
+};
+
+fn main() {
+    if let Err(error) = run() {
+        eprintln!("evidence-applicability: {error}");
+        std::process::exit(1);
+    }
+}
+
+fn run() -> Result<(), Box<dyn Error>> {
+    let mut bytes = Vec::new();
+    io::stdin()
+        .take((MAX_APPLICABILITY_QUERY_BYTES + 1) as u64)
+        .read_to_end(&mut bytes)?;
+    if bytes.len() > MAX_APPLICABILITY_QUERY_BYTES {
+        return Err(format!(
+            "applicability query exceeds the {MAX_APPLICABILITY_QUERY_BYTES}-byte limit"
+        )
+        .into());
+    }
+
+    let input = String::from_utf8(bytes)?;
+    let query: ApplicabilityQuery = serde_json::from_str(&input)?;
+    let evaluation = evaluate_query(&query)?;
+    serde_json::to_writer_pretty(io::stdout().lock(), &evaluation)?;
+    println!();
+    Ok(())
+}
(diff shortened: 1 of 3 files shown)


teamleaderleo:
Review: the shared applicability direction is strong and the CI failure is only rustfmt, but I see one semantic fail-closed blocker before this should become common substrate.

`EvidenceRequirements::default()` currently produces zero dimensions, and `evaluate_query()` therefore returns `APPLIES` vacuously.

For v1 that makes an omission bug dangerous:

```text
producer forgets every applicability constraint
-> requirements = {}
-> evaluator says APPLIES everywhere
```

That conflicts with the issue's north star that evidence should say where it is valid and with the wider Cultist rule that missing material applicability should not silently become a match.

Please require at least one explicit requirement (`repository`, `revision`, `work`, or `scope`) before evaluation. A future real use case for intentionally global evidence can earn an explicit global/unscoped marker rather than overloading absence of constraints.

Add a control roughly equivalent to:

```text
requirements = {}
context = anything
-> reject as malformed / underspecified
```

This also means the existing unknown-schema test should construct one valid requirement before changing `schema_version`, so it tests the intended failure rather than relying on an otherwise-invalid empty query.

The rest of the v1 boundary looks directionally right: exact dimensions, mismatch > missing, component-safe path prefix semantics, no authority/disposition coupling, and explicit dimension receipts. After this plus rustfmt, I’d rerun on current main because #121/#122 have moved underneath the original merge ref.

Refs #123 #115 #120 #122.

Issue #125, opened 2026-08-19
Research evidence roles for safe lossy agent projections


teamleaderleo:
## Thesis

C1 (#113/#114) shows that Cultist can compactly transport an `AnalysisReport` without losing meaning. Terse projection (#119/#121) asks a different question: which parts can be omitted from the *initial* agent view without changing the next correct action?

The first #121 experiment exposed a blocker in the current canonical model:

```rust
Evidence {
    message,
    location,
}
```

Evidence is not typed by its decision role. A renderer therefore cannot distinguish a verbose supporting receipt from the one evidence entry that carries a counterexample, platform limit, freshness boundary, or clearing condition.

A lossy renderer that drops all evidence prose can be much smaller and still be wrong.

## Research question

> What is the smallest evidence-role vocabulary that lets Cultist omit routine support while forcing decision-changing evidence to remain visible?

Do not change `AnalysisReport` schema in the first experiment.

## Candidate v0 roles

Start smaller than a universal ontology:

```text
support
  evidence that backs the claim but need not always appear in the initial projection

counterexample
  evidence that weakens/generalizes against the tempting conclusion

limit
  coverage/scope/platform/freshness/identity boundary that changes how far the claim may be used

clearing
  evidence/condition whose arrival would resolve a HOLD/UNKNOWN or reopen/advance a decision
```

`freshness` may later deserve its own temporal relation instead of living under `limit`; the experiment should tell us whether that distinction changes behavior.

## Why roles, not keywords

Do not classify evidence by words such as `only`, `linux`, `stale`, `counterexample`, or `limit` in arbitrary prose. The producer/analyzer that knows why an evidence record exists should emit the role.

If a producer cannot state the role deterministically, retain the evidence or keep the classification UNKNOWN.

## First adversarial fixtures

Build a research-only projection harness around representative `AnalysisReport`-like claims.

### A. Safe support omission

Canonical state:

```text
P claim: direct path overlap exists
support: current side changed src/auth.rs
support: other side changed src/auth.rs
```

Initial projection may omit support prose if the claim itself remains self-contained and expandable.

### B. Hidden platform boundary

```text
P claim: target test passed
limit: execution covered Linux only
```

Projection must not become an unqualified `PASS` or equivalent if the consumer might infer cross-platform evidence.

### C. Counterexample changes action

```text
O claim: 14/15 comparable changes update companion B
counterexample: the one omission is a docs-only change matching the current diff
```

[...]

PR #133, opened 2026-08-19
research: test existing report primitives as terse safety control

teamleaderleo:
Counter-hypothesis for #125 / #126.

## Question

Do we actually need a canonical `EvidenceRole` field to keep decision-changing information visible in terse output, or can producers already promote important material into existing `Claim` / `question` primitives?

## Controls

The integration fixture uses the actual merged `render_terse_analysis_report` and compares three hidden-vs-explicit representations.

### Coverage boundary

Hidden form:

```text
PROVEN target test passed
  evidence: execution covered Linux only
```

Explicit existing-primitives form:

```text
PROVEN target test passed
UNKNOWN cross-platform coverage is not established
```

The current terse renderer drops the hidden evidence but retains the explicit UNKNOWN.

### Counterexample

Hidden form keeps the matching reviewed exception inside generic Evidence.

Explicit form promotes it to a second OBSERVED claim. The current terse renderer retains it.

### Clearing step

Blocked form contains only the UNKNOWN. The explicit form uses the existing finding `question` field:

```text
Q Run exact target execution at HEAD?
```

The current terse renderer retains the question.

## What a positive result would mean

It would weaken the naive conclusion from #126:

```text
"we need EvidenceRole in the canonical report"
```

into the narrower result:

```text
"decision-changing information must be represented above generic receipt prose before lossy projection"
```

The remaining question would be whether existing Claim/question semantics are precise enough for machine consumers, or whether a smaller typed boundary/frontier vocabulary is still earned.

## Boundary

- test-only research fixture;
- no production schema/renderer changes;
- no claim that an OBSERVED counterexample claim carries an explicit `counterexample` role;
- no claim that free-text questions are a final machine action protocol;
- this is deliberately an alternative hypothesis to the role enum, not another proposal layered on top.

Refs #106 #109 #115 #119 #121 #125 #126.

### DIFF
--- a/tests/terse_existing_primitives.rs
+++ b/tests/terse_existing_primitives.rs
@@ -14,10 +14,12 @@ fn finding_with_hidden_limit() -> AnalysisReport {
         analysis: "terse-existing-primitives".to_string(),
         repository: "/repo".to_string(),
         claims: Vec::new(),
-        findings: vec![Finding::new("execution", "Target execution").with_claim(
-            Claim::new(ClaimKind::Proven, "target test passed")
-                .with_evidence(Evidence::new("execution covered Linux only")),
-        )],
+        findings: vec![
+            Finding::new("execution", "Target execution").with_claim(
+                Claim::new(ClaimKind::Proven, "target test passed")
+                    .with_evidence(Evidence::new("execution covered Linux only")),
+            ),
+        ],
     }
 }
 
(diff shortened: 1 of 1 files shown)


teamleaderleo (commented):
This is a useful counter-hypothesis to #126: it narrows the result from “canonical EvidenceRole is required” to “decision-changing material has to live above generic receipt prose before lossy projection.” I’d keep that distinction.

One follow-up control would make the remaining gap especially crisp: build two explicit-claim reports with the same claim kinds and similar topology where one second `OBSERVED` claim is ordinary support and the other is a current-matching counterexample. Terse can preserve both strings, but the machine representation still gives the receiver no typed way to know which one should reverse/escalate the next action without interpreting prose.

That would leave us with a cleaner three-step result:

```text
generic Evidence only
  -> current terse can erase decision-changing content

promote into Claim / question
  -> current terse keeps the content

same ClaimKind, different decision role
  -> machine consumer still needs prose interpretation or a later earned role/frontier distinction
```

So I like this PR as pressure against premature schema growth; I’d avoid letting a green result drift into “existing primitives solve safe lossy projection” until that last ambiguity is exercised.

Issue #134, opened 2026-08-19
Research explicit ambient-context binding for compact agent envelopes


teamleaderleo:
## Thesis

Compact/terse Cultist messages can rely on ambient repository context only when that context is **explicitly bound and checkable**.

#115 proposes saying repository/revision/task once and then using relative references. #131 adds an exact `AnalysisReport` fingerprint through canonical C1. These are complementary, but not the same identity.

A report fingerprint means:

```text
these exact report semantics/order under the C1 codec
```

It does not automatically mean:

```text
this report was computed for the receiver's current HEAD/task/work item
```

because `AnalysisReport` does not universally carry every ambient input such as current revision, task identity, provider snapshot, or working-tree generation.

## Research question

> What is the minimum explicit context header needed so a compact/terse packet can safely omit repeated repository/revision/task narration while remaining independently reconstructible after agent replacement?

## First discriminator

Use the real #131 report fingerprint and a test-local external context:

```text
report R
context A = repo X, revision abc, task T1
context B = repo X, revision def, task T1
```

If the report bytes are unchanged:

```text
fingerprint(R@A.report) == fingerprint(R@B.report)
```

while the ambient applicability differs.

This proves report identity alone cannot establish revision freshness.

Then bind the report fingerprint to an explicit context identity and show:

```text
context-bound(A) != context-bound(B)
```

The first experiment should avoid inventing a production hash/grammar for the outer envelope. The property is the important part.

## Candidate context dimensions

Keep them optional/required by the consumer instead of one universal giant header:

```text
repository identity
revision / working-tree identity
task or work identity
provider / active-work snapshot identity
selected schema / projection version
base report/envelope fingerprint
```

#123's applicability model should decide which dimensions are required for a particular evidence use. A local repository scan may need fewer dimensions than an active-work review packet.

## Important distinction

Keep three identities separate:

```text
report fingerprint
  exact typed report semantics

ambient context
  repo/revision/work/provider coordinates in which the report is being used

applicability
  whether the admitted evidence requirements match the current context
```

Do not silently bake all context into every claim or make a report fingerprint mean more than its encoded bytes.

## Terse transport consequence

A terse payload such as:

```text
F1 preflight-overlap
  C1 P both sides modify src/auth.rs
```

may omit repository/revision/task only when the containing session/envelope proves those coordinates.

If the receiver lacks that context, the correct behavior is:

```text
CONTEXT_MISSING / UNKNOWN
-> request header/full envelope
```

[...]

Issue #127, opened 2026-08-19
Research stable evidence identity before delta-only agent packets


teamleaderleo:
## Thesis

#115 proposes delta messages once sender and receiver share prior Cultist state. #119/#121 already use report-local refs such as `F1` as a deliberately temporary convenience.

Before a delta protocol, Cultist needs to distinguish **position** from **identity**.

A positional reference is cheap inside one immutable report:

```text
F1
F2
```

but it is unsafe across recomputation. If finding order changes, `F1` can silently refer to a different finding even when both findings still exist.

## Research question

> What minimum identity/fingerprint contract lets Cultist send deltas against shared evidence without applying a transition to the wrong semantic object?

## First counterexample

Start with two findings:

```text
E1 = [A, B]
```

A recomputation changes ranking/order only:

```text
E2 = [B, A]
```

The semantic set is unchanged.

A positional delta:

```text
resolve F1
```

means `resolve A` under E1 and `resolve B` under E2.

The same bytes therefore have two incompatible meanings unless the exact base fingerprint is bound and checked.

### Report snapshot identity

Fingerprint the exact immutable report/envelope bytes or canonical semantic encoding.

Use it to require:

```text
delta.base == receiver.current_snapshot
```

Mismatch fails closed and requires a full packet/rebase.

### Object snapshot identity

Content-address an exact claim/finding/evidence object when useful.

This survives reordering but changes when the object content changes.

### Semantic lineage identity

A later problem: whether two changed objects are revisions of the *same* finding/decision/question.

Do not infer lineage merely because titles/messages look similar. Analyzer-owned stable keys may eventually use exact target/discriminator identities where semantics support them.

These are different jobs; do not force one ID to solve all three.

## First research harness

Build a tiny deterministic fixture showing:

1. positional refs are valid inside one exact snapshot;
2. applying a positional delta to the wrong/reordered base changes the wrong object;
3. binding the delta to an exact base fingerprint rejects the mismatch;
4. a stable object key can survive pure reordering in the synthetic fixture;
5. changing an object's semantic key forces a new identity instead of silently carrying old state.

No product schema change in v0.

## Candidate delta envelope

Conceptually:

```text
D1
base <exact-report-fingerprint>
+ <object>
- <object-id>
~ <old-object-id> <new-object>
```

Syntax is not the research goal yet.

The key invariant is:

> no delta is interpreted unless the receiver proves it is applying the transition to the exact state the sender referenced.

## Composition with C1

[...]

PR #135, opened 2026-08-19
research: prove compact envelopes need ambient context binding

teamleaderleo:
Starts #134 as a stacked research control on top of #131.

## Question

Does an exact `AnalysisReport` fingerprint also prove that the receiver is using the report in the same repository/revision/task context?

No. This fixture makes that boundary executable.

## Real fingerprint + external context

The test imports the actual #131 C1/SHA-256 report fingerprint implementation.

It constructs one report whose typed semantics do not contain a Git revision, then evaluates it under two external contexts:

```text
repo teamleaderleo/cultist
revision aaaaaaaa
```

and:

```text
repo teamleaderleo/cultist
revision bbbbbbbb
```

The report fingerprint remains identical because the report bytes/semantics are identical.

A tiny test-local `BoundEnvelopeIdentity` then binds only context dimensions declared required by the consumer. Requiring revision makes the two bound identities differ; omitting required revision fails with `MissingRevision` instead of guessing.

## Negative-overbinding control

When the fixture requires only repository identity, changes to revision/task are deliberately absent from the bound identity and do not change it. This tests the JEI side too: do not force every possible context coordinate into every message.

## What this proves if green

```text
report fingerprint
  != repository-state freshness
```

and:

```text
relative terse shorthand is safe only inside an envelope/session that proves the context dimensions required for that use
```

## Boundary

- test-only outer binding model;
- no production envelope grammar/hash;
- no duplicate applicability evaluator;
- no claim that every report requires revision/task/provider coordinates;
- semantic lineage remains separate from report snapshot and ambient context identity.

This PR is stacked on `research/c1-report-fingerprint`; it should be rebased/retargeted if #131 lands.

Refs #115 #119 #123 #127 #131 #134.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,3 +14,6 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 syn = { version = "2.0", features = ["full", "visit"] }
 walkdir = "2.5"
+
+[dev-dependencies]
+sha2 = "0.11"
--- a/examples/report_fingerprint.rs
+++ b/examples/report_fingerprint.rs
@@ -1,0 +1,42 @@
+#![allow(dead_code)]
+
+#[path = "../src/compact_ir.rs"]
+mod compact_ir;
+#[path = "../src/finding.rs"]
+mod finding;
+#[path = "../src/report_fingerprint.rs"]
+mod report_fingerprint;
+
+use std::error::Error;
+use std::io::{self, Read};
+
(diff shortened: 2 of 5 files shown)


teamleaderleo:
Superseded by clean replay #136. #131 was rebased and merged while this stacked carrier was being created, so this PR retained superseded fingerprint history in its diff. #136 replays the single context-binding fixture directly on merged main; use that as the canonical experiment.

PR #136, opened 2026-08-19
research: replay ambient context binding on merged fingerprint

teamleaderleo:
Clean replay of #134 on current `main` after #131 merged. Supersedes the stale stacked carrier #135.

## Question

Does an exact `AnalysisReport` fingerprint also prove that the receiver is using the report in the same repository/revision/task context?

The fixture uses the now-merged C1/SHA-256 fingerprint implementation and keeps the ambient context outside the report.

## Controls

- one exact report is evaluated under revision `aaaaaaaa` and `bbbbbbbb`;
- its report fingerprint remains identical because report semantics/bytes are unchanged;
- a test-local bound envelope identity that *requires* revision changes across those contexts;
- missing required revision returns `MissingRevision` instead of guessing;
- when only repository identity is required, unrelated revision/task changes are deliberately omitted and do not perturb the bound identity.

## Meaning if green

```text
report fingerprint
  = exact report snapshot identity

ambient context binding
  = coordinates required for this use of the report
```

A compact/terse packet may omit repeated repo/revision/task narration only after its containing envelope/session proves the dimensions required for that consumer.

## Boundary

- test-only outer binding model;
- no production envelope grammar/hash;
- no duplicate applicability evaluator;
- no claim that every report needs every coordinate;
- semantic lineage remains separate.

Refs #115 #119 #123 #127 #131 #134 #135.

### DIFF
--- a/tests/ambient_context_binding.rs
+++ b/tests/ambient_context_binding.rs
@@ -1,0 +1,247 @@
+#[allow(dead_code)]
+#[path = "../src/compact_ir.rs"]
+mod compact_ir;
+#[allow(dead_code)]
+#[path = "../src/finding.rs"]
+mod finding;
+#[allow(dead_code)]
+#[path = "../src/report_fingerprint.rs"]
+mod report_fingerprint;
+
+use finding::{AnalysisReport, Claim, ClaimKind, REPORT_SCHEMA_VERSION};
+use report_fingerprint::{ReportFingerprint, fingerprint_report};
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+struct AmbientContext<'a> {
+    repository: Option<&'a str>,
+    revision: Option<&'a str>,
+    task: Option<&'a str>,
+}
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+struct ContextRequirements {
+    repository: bool,
+    revision: bool,
+    task: bool,
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+struct BoundEnvelopeIdentity {
+    report: ReportFingerprint,
+    repository: Option<String>,
+    revision: Option<String>,
+    task: Option<String>,
+}
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+enum BindError {
+    NoRequirements,
+    MissingRepository,
+    MissingRevision,
+    MissingTask,
+}
+
+fn bind_context(
+    report: &AnalysisReport,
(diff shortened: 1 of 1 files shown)


teamleaderleo (commented):
Rook 🦉 review of exact head `933085b8c0ee4ebf66045f4adf7889be9e41c532`.

No code-correctness blocker found in this ambient-context binding proof.

The report fingerprint stays an identity for the report snapshot itself, while required repository/revision/task dimensions live in a separate bound envelope. Required missing values fail closed; unrequired dimensions are omitted rather than allowed to perturb identity. The revision-movement control therefore changes the bound identity while preserving the underlying report fingerprint, which is the intended discriminator.

The first test's equal-fingerprint assertion is intentionally trivial-looking because both calls fingerprint the same report and the ambient context lives outside it; the following bound-identity test is the meaningful control that proves external revision movement is invisible to report identity but visible when revision is required by the envelope.

Exact head CI `32241509533` and generated provenance `32241509539` are green.

Promotion limit is scope/age: this is a test-only outer binding model from 2026-08-19. Current product work now has shared applicability and provider-current gates; any public envelope should reuse those current primitives and retain this principle rather than promote this test-local binding type verbatim.

teamleaderleo (commented):
Flint 🐺 independent/current-head review of exact head `1dd949a89bb2780b551d80e4c7d845cb16186d41` on `main@e1216c6a311d7f98e0d2a93724d7e8216c56f53d`.

No code-correctness blocker found in this test-only ambient-context discriminator.

I attacked the identity/applicability separation directly:

- the same typed `AnalysisReport` keeps the same C1/SHA-256 report fingerprint when only external revision context moves;
- when revision is explicitly required by the containing context identity, the bound identity changes across `aaaaaaaa -> bbbbbbbb` while the underlying report fingerprint stays equal;
- missing required revision fails closed as `MissingRevision` rather than being guessed from cwd/chat/branch state;
- empty requirements fail closed instead of manufacturing context-free/global applicability;
- when only repository is required, revision/task movement is omitted from the bound identity and does not perturb it;
- report fingerprint and ambient repository/revision/task values remain separately inspectable fields.

The local `ContextRequirements`/`bind_context` type is a research identity-construction fixture, not a replacement for shared #123 current-vs-required applicability evaluation. Rook's earlier promotion limit still applies: any public envelope should use current shared applicability/context primitives rather than promote this local type verbatim.

Hosted validation is green on this exact one-file current-main replay: CI `32744115925` and generated provenance `32744115948` both succeeded. Main remained at the exact base through review.

Issue #137, opened 2026-08-19
Product pressure test: optimize for decision-changing repository evidence


teamleaderleo:
## Thesis

Cultist's strongest product slices are the ones that change what a worker should do next before the worker burns time or ships the wrong thing.

Examples already in the repository include:

- generated-companion evidence that can send a worker to regenerate an output;
- CI selector evidence that can expose a green command selecting zero intended tests;
- concurrent-work/preflight evidence that can trigger coordination before duplicated or conflicting work advances;
- bounded target history that can surface an earned invariant before a fresh worker repeats an old mistake.

The broader repository-memory thesis still looks valuable. This issue proposes a sharper product pressure test while the existing research lanes continue:

> Prefer evidence that changes the next justified action, and measure whether Cultist reduces manual rediscovery and expensive wrong turns.

This is a prioritization/evaluation lens, not a request to freeze C1, JEI, review, applicability, decision-memory, or other research.

## Why open this now

Cultist has accumulated increasingly careful semantics around provenance, uncertainty, applicability, compact representation, active-work identity, and durable memory. That discipline is useful. The next complementary discriminator is behavioral:

```text
repository + task
-> worker investigates/edits

versus

repository + task + selected Cultist evidence
-> worker investigates/edits
```

The question is whether the second worker reaches decisive evidence earlier, avoids a known failed approach, inspects fewer irrelevant surfaces, or coordinates sooner.

The existing agent-context replay already identified this as the important next gate after proving the packet can recover useful history.

## Candidate product principle

A finding/evidence family earns prominent automatic delivery when it repeatedly does at least one of:

- changes the next action;
- prevents an expensive wrong turn;
- exposes a missing validation step;
- invalidates evidence the worker was about to rely on;
- reveals active work that requires coordination;
- surfaces reviewed rationale whose absence would cause repeated investigation;
- turns a material `UNKNOWN` into a bounded next discriminator.

Interesting-but-low-consequence observations can remain queryable, research-only, or lower-priority output.

This keeps the existing epistemic vocabulary. `PROVEN / DERIVED / OBSERVED / INFERRED / UNKNOWN` answers what the evidence establishes. Action relevance answers why the worker should spend attention on it now. These are separate axes.

## First behavioral experiment

Use held-out tasks with known earned invariants or known failure paths.

For each task, compare fresh workers under two conditions:

### Control

```text
checkout + task
```

### Cultist-assisted

```text
checkout + task + bounded pre-edit evidence
```

[...]

PR #139, opened 2026-08-19
research: enforce semantic byte budget on agent packet

teamleaderleo:
Progresses #67 and composes with #62/#106/#115.

## What

Promote the existing count-bounded `agent_context_packet` research object from measurement-only schema v4 to a first **hard semantic byte-budget** experiment (schema v5).

The packet now records:

```text
budget.max_serialized_bytes
candidate_serialized_bytes
serialized_bytes
semantic_evictions[{class,count}]
```

The default research target is 32 KiB and can be overridden with `CARGO_CULTIST_PACKET_MAX_BYTES`.

## Deterministic eviction policy

Protected core is never silently removed:

- packet/repository/target identity;
- direct evidence;
- applicable guidance identities;
- reviewed decisions;
- material UNKNOWNs;
- budget and omission receipts.

Evictable detail is removed one unit at a time in this order:

```text
1. historical support examples, lowest-ranked relation first
2. historical counterexamples
3. lowest-ranked historical companion relation
4. oldest recent-history summary
5. companion-exclusion detail
```

Removing support/counterexample examples increments the existing relation-local `examples_omitted` / `counterexamples_omitted` algebra. Every semantic eviction also increments a typed packet-level receipt.

If no evictable detail remains and the protected core still exceeds the declared budget, compilation fails with `ProtectedCoreTooLarge` instead of hiding guidance, decisions, or UNKNOWNs.

There is no byte slicing/post-serialization truncation.

## Measurement boundary

The compiler measures the full candidate object first, then selected output:

```text
candidate_serialized_bytes
selected serialized_bytes
max_serialized_bytes
```

JSON is reserialized after deterministic evictions in this research slice. Optimization is deferred until the policy earns itself.

## Adversarial controls

The ordinary `cargo test` suite now includes the real example module and uses long commit subjects to require:

- support examples are evicted before counterexamples;
- final selected JSON is valid and <= the declared budget;
- candidate bytes exceed selected bytes when semantic eviction occurs;
- protected direct evidence/guidance/UNKNOWNs survive;
- identical candidate + budget yields byte-identical output;
- an impossible protected core fails explicitly.

## Boundary

This is one policy discriminator, not the final universal JEI ranking. #125/#126/#133 are still testing whether decision-changing evidence must be promoted above generic `Evidence` receipts before lossy selection. Transport/admission bounds from C1/#118 remain independent and still fail rather than evict.

No primary CLI behavior changes; this is the existing research `agent_context_packet` example plus its standard test harness.

Refs #62 #67 #73 #106 #115 #118 #125 #126 #131 #133.

### DIFF
--- a/examples/agent_context_packet.rs
+++ b/examples/agent_context_packet.rs
@@ -379,7 +379,9 @@ fn record_semantic_eviction(packet: &mut AgentContextPacket, class: &'static str
         receipt.count += 1;
         return;
     }
-    packet.semantic_evictions.push(SemanticEviction { class, count: 1 });
+    packet
+        .semantic_evictions
+        .push(SemanticEviction { class, count: 1 });
 }
 
 fn git_repo_root(path: &Path) -> Result<PathBuf, Box<dyn Error>> {
@@ -674,7 +676,10 @@ mod tests {
                     support: 9,
                     opportunities: 10,
                     support_percent: 90.0,
-                    examples: vec![summary("high-support-a", 900), summary("high-support-b", 900)],
+                    examples: vec![
+                        summary("high-support-a", 900),
+                        summary("high-support-b", 900),
+                    ],
                     examples_omitted: 7,
                     counterexamples: vec![summary("high-counter", 900)],
                     counterexamples_omitted: 0,
@@ -696,7 +701,8 @@ mod tests {
                 changed_paths: 101,
             }],
             unknowns: vec![
-                "Material UNKNOWN remains protected across semantic budget compilation.".to_string(),
+                "Material UNKNOWN remains protected across semantic budget compilation."
+                    .to_string(),
             ],
             truncation: Vec::new(),
         }
@@ -734,7 +740,10 @@ mod tests {
         let json: serde_json::Value = serde_json::from_str(&rendered).unwrap();
 
         assert!(rendered.len() <= max);
-        assert_eq!(json["serialized_bytes"].as_u64().unwrap() as usize, rendered.len());
+        assert_eq!(
+            json["serialized_bytes"].as_u64().unwrap() as usize,
+            rendered.len()
+        );
         assert!(
             json["candidate_serialized_bytes"].as_u64().unwrap()
                 > json["serialized_bytes"].as_u64().unwrap()
--- a/examples/agent_context_packet.rs
+++ b/examples/agent_context_packet.rs
@@ -1,6 +1,7 @@
 use std::collections::{BTreeMap, BTreeSet};
 use std::env;
 use std::error::Error;
+use std::fmt;
 use std::fs;
 use std::path::{Path, PathBuf};
 use std::process::Command;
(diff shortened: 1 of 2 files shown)


teamleaderleo (commented):
Skeptical byte-budget read: the hard-budget mechanism looks coherent, but the first eviction policy can still spend decision-changing counterevidence before cheaper lower-ranked context.

`historical_companions` is sorted strongest-first by support. `evict_one_semantic_detail()` walks it in reverse for support examples, then again in reverse for counterexamples, and only after **all** counterexample detail is gone does it drop the lowest-ranked companion relation / old history.

That creates an adversarial case worth pinning before this policy earns promotion:

```text
high-ranked relation 9/10
  one counterexample whose exact content matches the current task

low-ranked relation 5/10
  ordinary support/counterexample detail

budget pressure
  only one of {high-ranked decisive counterexample, whole low-ranked relation} can remain
```

The current policy can evict the high-ranked counterexample text while retaining the lower-ranked relation. `counterexamples_omitted += 1` keeps an omission receipt, which is good, but the omitted content may be exactly what reverses the tempting conclusion. That composes directly with #125/#126/#133 and the #137 product-pressure question.

I would add an adversarial control before treating the order as a useful JEI candidate. Either:

- prove the current ordering still preserves the modeled next action in such a fixture; or
- change the ordering so lower-ranked whole relations / older broad context are eligible before counterexample detail from stronger relations; or
- explicitly defer counterexample eviction until the evidence-role/current-task discriminator exists.

This does not argue against the hard byte-budget experiment. It is the kind of negative control that can tell us whether semantic eviction is actually safer than count-only truncation.

teamleaderleo:
Review blocker before this becomes the first hard semantic byte budget: the current eviction ladder can remove **exact companion exclusion receipts** and retain only `semantic_evictions: { class: "companion_exclusion_detail", count: N }`.

That regresses a boundary we already earned in the #62/SmolRunner packet replay. Exact excluded commits + reasons were added specifically because a count alone can make historical support look stronger than the admitted cohort warrants: consumers need to know *which* history was rejected and *why*, not merely that some provenance disappeared.

Current order is good in one respect—support examples, counterexamples, companion relations, and recent history are evicted before exclusions—but the final `companion_exclusions.pop()` still lets a sufficiently tight budget erase the cohort receipt entirely.

For v5 I would prefer one of these fail-closed options:

1. make exact `companion_exclusions` part of the protected core, so a budget smaller than target/guidance/decisions/UNKNOWNs/**cohort exclusions** fails explicitly; or
2. compact an exclusion before eviction into a typed exact receipt that still retains at least commit SHA + reason (and any field required to interpret the broad/revert/mechanical cohort), with the byte-saving transformation itself receipted.

I would **not** reduce exclusions to only class/count.

Please add a discriminator where the budget is tight enough to remove all examples/relations/history but still requires the exact excluded commit identity/reason to survive, plus an impossible-budget case after that protected/compact exclusion receipt is accounted for.

The rest of the design looks promising: byte count is actual UTF-8 serialized length, encode-time eviction is deterministic, direct evidence/guidance/reviewed decisions/UNKNOWNs are protected, and candidate/final byte receipts are useful. This is one provenance boundary worth protecting before promotion.

Refs #62 #67 #106 #139.

Issue #144, opened 2026-08-19
Research durable UNKNOWN obligations and clearing conditions


teamleaderleo:
## Thesis

Cultist already treats UNKNOWN as useful output. JEI/review/lifecycle work needs a stronger handoff primitive:

> when investigation ends with one material missing discriminator, preserve exactly what is known, what remains missing, and what evidence would clear or reopen the obligation.

This lets a later worker resume at the frontier instead of repeating completed archaeology.

## Candidate object

```text
Obligation
  id
  question
  subject / applicability requirements
  established evidence refs
  missing discriminator
  clearing condition(s)
  invalidation / reopen condition(s)
  source work / revision
```

Example:

```text
U17
question
  does target T pass at current HEAD?

established
  relevant guidance recovered
  history inspected
  previous Linux receipt exists but applies to old head

missing
  exact-head target execution

clears_when
  test_result(target=T, revision=abc123)

reopens_when
  revision != abc123
```

## First fixtures

1. exact-head test execution obligation;
2. provider-current work identity missing;
3. decision applicability missing current target coordinate;
4. generator ownership UNKNOWN with one deterministic ownership probe;
5. obligation that becomes obsolete because the target/work itself disappears or is superseded.

## Requirements

- deterministic/local v0;
- applicability evaluated by shared #123/#124 semantics;
- stable identity bound to an exact snapshot before cross-run deltas (#127/#128);
- clearing condition names observable evidence, not model prose;
- reopening is explicit when previously clearing evidence expires;
- solved obligations remain inspectable as history when useful;
- no mutation or authority implied by clearing an epistemic obligation.

## Composition

This should feed:

- #62/#106 pre-edit JEI;
- #74 lifecycle handoff;
- #109 review frontier;
- #115 compact IR transitions;
- #119 terse NEXT/HOLD-style projections;
- decision memory where accepted rationale closes a recurring question;
- future evidence acquisition planning.

## Success criteria

- one worker can stop with a durable obligation and a fresh worker can resume without repeating already-recorded investigation;
- one exact new evidence object clears the obligation deterministically;
- a moved coordinate reopens the obligation;
- missing context remains UNKNOWN rather than treated as satisfied;
- compact projection can transmit the obligation without losing the clearing condition;
- the object remains useful without an LLM.

North star:

> Preserve the edge of repository knowledge: what remains unknown, what would answer it, and when the answer must be reconsidered.

[...]

Issue #141, opened 2026-08-19
Research justification graphs and truth maintenance for repository evidence


teamleaderleo:
## Thesis

Cultist now has exact coordinate applicability (#123/#124), stable snapshot identity research (#127/#128), evidence-role research (#125), and multiple consumers that need to react when evidence moves.

The next question is deeper than freshness:

> When one evidence object becomes invalid, which claims, obligations, review dispositions, or decisions actually need to change — and which remain justified by independent evidence?

A whole packet should not become stale merely because one receipt moved. Conversely, a claim should not survive merely because another unrelated receipt is still present.

## Research question

Can Cultist represent a small deterministic justification graph over existing evidence objects so that claim state can be recomputed after evidence invalidation without feature-specific stale flags?

## Candidate v0 relations

Start deliberately small:

```text
supports
  evidence independently supports a claim

counters
  evidence weakens or defeats the tempting conclusion

limits
  evidence narrows scope/platform/revision/applicability

depends_on
  one derived object requires another object to remain valid

clears
  evidence satisfies a previously explicit obligation / UNKNOWN
```

These are research relations, not a frozen canonical vocabulary.

### A. Single support invalidates claim

```text
E1 exact-head target test passed
E1 -> supports C1

head moves
-> E1 invalid
-> C1 becomes unsupported / UNKNOWN
```

### B. Independent support survives

```text
E1 -> supports C1
E2 -> supports C1

E1 invalid
E2 still applicable
-> C1 remains supported
```

### C. Limit invalidates stronger projection

```text
E1 target test passed
E2 Linux-only execution limit
E1 supports C1
E2 limits C1
```

A projection that drops E2 must not silently strengthen C1.

### D. Counterexample changes cohort interpretation

```text
E1 14/15 companion support
E2 docs-only counterexample matching current change
```

Removing E2 changes the justified next question and should be visible.

### E. Clearing evidence resolves durable UNKNOWN

```text
U1 exact-head target execution missing
E3 exact-head target test result
E3 clears U1
```

## Implementation slice

Keep v0 research-only and reuse current report/applicability objects where practical.

Candidate types:

```text
EvidenceNodeId
ClaimNodeId
JustificationEdge
JustificationRelation
JustificationGraph
ReevaluationReceipt
```

Requirements:

- deterministic ordering;
- stable explicit node identity inside one exact research snapshot;
- applicability evaluation delegated to the shared evaluator where coordinates are present;
[...]

Issue #10, opened 2026-08-18
Add explicit decision records: teach Cultist why an exception exists


teamleaderleo:
## Idea

A suppression should be able to preserve rationale instead of merely silencing a finding.

Cultist could let maintainers record an intentional exception as project memory:

```text
cargo cultist teach <finding> --reason "This test needs a stable pathname for a subprocess"
```

The exact UX/file format is open for exploration.

## Possible decision record

```toml
[[decisions]]
kind = "temp-file-precedent"
scope = "tests/subprocess.rs"
reason = "The child process reopens this pathname after creation."
introduced_by = "PR #123"
expires = "2027-06-01"
```

Possible conditional expiry could eventually reference dependency/toolchain versions.

## Desired behavior

Future findings should be able to report:

```text
KNOWN EXCEPTION
Reason: ...
Recorded in: ...
Still applicable: unknown / yes / condition appears expired
```

A suppression or decision with no rationale can itself become a knowledge-debt finding.

## Design questions

- repo-local file vs inline annotation vs both
- stable finding identity across refactors
- scope matching
- expiry conditions
- reviewability in Git

## Success criteria

- A human can resolve a recurring Cultist question by recording why.
- The explanation lives in version control and is visible to future analysis.
- Decisions do not silently suppress unrelated future code.

PR #156, opened 2026-08-19
research: add justification graph semantics

teamleaderleo:
Progresses #141 with a bounded truth-maintenance experiment over current applicability semantics.

## V0 graph

```text
EvidenceNode
  -> ClaimNode
     support | counterexample | limit | dependency

EvidenceNode
  -> ObligationNode
     clearing
```

Every evidence node carries the existing `EvidenceRequirements` contract and delegates current-coordinate evaluation to the shared applicability evaluator.

## Semantics retained

- independent support survives when one support receipt becomes invalid;
- sole invalidated support returns the claim justification to `unknown`;
- counterexample and limit receipts remain explicit;
- hard v0 dependencies with INVALID/UNKNOWN applicability block supported projection;
- an obligation can exist OPEN before any clearing receipt is observed;
- exact current clearing evidence resolves an obligation and revision movement reopens it;
- reevaluation reports downstream claim/obligation targets whose evidence applicability changed.

The graph remains bipartite in v0. Claim-to-claim derivations and cycles stay outside this carrier.

The original four semantic blobs were repeatedly reanchored as unrelated main work landed. Final exact-current carrier:

```text
main: eb9c0eeed395b9e40addc30c408d4ff83f24ab42
head: e33cc0e736d3d9092cd23782cef4a8ad2cd1935a
commits: 1
files: 4
```

Ordinary CI:

```text
run: 32272350023
job: 96131568746
result: success
```

Generated-provenance review:

```text
run: 32272350050
job: 96131568956
result: success
```

Formatter, strict Clippy, project/review/closure/external-reference guards, active-work preflight, full tests, and repository/history/CI-filter/diff dogfood all passed. Main remained at the exact base through the final gate.

The earlier semantic receipt remains in `research/justification-graph.md`; this current-main run proves the same graph semantics compose with the modern repository.

## Boundary / next

Research only; no product CLI/report-schema change and no automatic truth authority.

The next earned child is #159: durable UNKNOWN obligations that preserve the missing discriminator before any clearing receipt exists.

Files: `src/justification.rs`, `tests/justification.rs`, `examples/justification_graph.rs`, `research/justification-graph.md`.

Refs #6 #10 #62 #74 #106 #109 #115 #123 #125 #127 #141 #144 #159.

### DIFF
--- a/examples/justification_graph.rs
+++ b/examples/justification_graph.rs
@@ -1,0 +1,65 @@
+#![allow(dead_code)]
+
+use std::error::Error;
+use std::io::{self, Read};
+
+use serde::Deserialize;
+
+#[path = "../src/applicability.rs"]
+mod applicability;
+#[path = "../src/justification.rs"]
+mod justification;
+
+use applicability::EvaluationContext;
+use justification::{JustificationGraph, evaluate_graph, reevaluate_graph};
+
+const MAX_INPUT_BYTES: u64 = 1024 * 1024;
+
+#[derive(Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
+struct EvaluationInput {
+    graph: JustificationGraph,
+    context: EvaluationContext,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
+struct ReevaluationInput {
+    graph: JustificationGraph,
+    before_context: EvaluationContext,
+    after_context: EvaluationContext,
+}
+
+fn main() -> Result<(), Box<dyn Error>> {
+    let args = std::env::args().skip(1).collect::<Vec<_>>();
+    let mut input = Vec::new();
+    io::stdin()
+        .take(MAX_INPUT_BYTES + 1)
+        .read_to_end(&mut input)?;
+    if input.len() as u64 > MAX_INPUT_BYTES {
+        return Err("justification input exceeds 1 MiB".into());
+    }
+
+    match args.as_slice() {
+        [] => {
+            let request: EvaluationInput = serde_json::from_slice(&input)?;
+            let evaluation = evaluate_graph(&request.graph, &request.context)?;
+            serde_json::to_writer_pretty(io::stdout().lock(), &evaluation)?;
+        }
+        [flag] if flag == "--reevaluate" => {
+            let request: ReevaluationInput = serde_json::from_slice(&input)?;
+            let receipt = reevaluate_graph(
+                &request.graph,
+                &request.before_context,
+                &request.after_context,
+            )?;
+            serde_json::to_writer_pretty(io::stdout().lock(), &receipt)?;
+        }
+        _ => {
+            return Err("usage: cultist_justification [--reevaluate] < request.json".into());
+        }
+    }
+
+    println!();
+    Ok(())
+}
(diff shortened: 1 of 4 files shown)


teamleaderleo:
## Cross-lane discriminator from #125/#133: prove relation role is not intrinsic to Evidence

This graph model is a cleaner answer to the earlier `EvidenceRole` temptation because `support / counterexample / limit` are represented as **edges from evidence to a target**, not properties of the evidence node itself.

One adversarial control would make that distinction executable:

```text
E1
  -> support C1
  -> counterexample C2
```

or:

```text
E1
  -> support C1
  -> limit C2
```

with the same exact `EvidenceNode` and applicability status.

Expected result:

- graph validates both edges;
- C1's relation receipt lists E1 under support;
- C2 lists the same E1 under counterexample/limit;
- changing E1 applicability updates both downstream targets through their distinct relation buckets.

Why this is worth retaining: if the same evidence can legitimately play different roles against different claims, then a canonical `Evidence.role` field would be conceptually wrong, not merely redundant.

That would turn #133's current conclusion—"new EvidenceRole enum not yet earned"—into a stronger positive reason for this relational model.

No request to make counterexamples automatically negate claims; keeping the relation explicit without flattening it is the useful property.

teamleaderleo:
Dogfood from the immediately stacked #144 durable-UNKNOWN experiment exposed one useful v0 boundary here: `validate_graph()` currently requires every obligation to already have a `clearing` edge. A durable open obligation can legitimately exist before any clearing evidence object has arrived, so a future revision of #141 should either allow zero-edge open obligations or keep prospective clearing conditions outside the observed-evidence graph. The #144 carrier takes the latter path for now and preserves this as a counterexample rather than manufacturing a future evidence node. This does not affect the current claim-support controls; it narrows what #141's obligation node can claim to represent.

teamleaderleo:
## Semantic status guard: `Supported` must not become a shortcut around counterexamples/limits

The current evaluator deliberately keeps these separate:

```text
status = Supported
counterexamples.applies = [E2]
limits.applies = [E3]
```

when at least one support applies and dependencies are not blocked. I agree with the non-collapsing model: a counterexample should not automatically negate support, and a limit should not erase the supported core.

The risky part is the **name/consumer contract**. A downstream projection can easily do:

```text
if status == supported { proceed }
```

and silently ignore the decision-changing receipts that #125/#126 proved can matter.

Useful adversarial pair:

```text
Graph A
  support E1 -> C1
  => status Supported

Graph B
  support E1 -> C1
  counterexample E2 -> C1
  limit E3 -> C1
  => status Supported
```

If a minimal consumer that reads only `status` takes the same action for A/B while a receipt-aware consumer does not, the graph has reproduced the lossy-projection failure at a higher level.

Possible remedies to test, not prescriptions:

- rename the status to something narrower such as `has_applicable_support` / `support_present`;
- require a projection helper to expose `supported + contested/limited` instead of encouraging raw status branching;
- keep `Supported` but add an explicit regression/docs invariant: **Supported is not unqualified, accepted, safe, or actionable without relation receipts.**

This is the same semantic-strengthening failure class that killed terse `OK` and challenged the `check` alias. The graph itself is good; I want the one-word projection to be harder to misuse than the full receipts.

PR #138, opened 2026-08-19
Add pinned external dogfood case registry

teamleaderleo:
## Goal

Turn external dogfood from one-off commands into a small accumulating queue of exact replay coordinates.

This builds on merged #129 and keeps the same rule: large external corpus work stays manual/on-demand while normal CI remains local and cheap.

## Registry

`research/external-dogfood-cases.json` introduces three admitted states:

- `replayable` — exact repository/revision coordinates plus today's Cultist probes can evaluate something useful;
- `adapter_gap` — the case is pinned, but the interesting claim needs evidence today's product does not ingest yet;
- `needs_pin` — available for useful historical stories whose exact replay coordinate still needs recovery.

Registered replay cases require exact 40-hex `ref` and exact 40-hex `base` when a base is used. PR/issue numbers remain provenance labels; the executable identity is repository + exact revisions + evidence target.

### SmolRunner: replayable earned-history control

```text
teamleaderleo/smolrunner@ed3b70e375a57eabce26f2311f798f75b33bdeb0
src/disposable_clone_runtime.rs
```

This preserves the #62 / #129 history discriminator.

### Cloud Hypervisor #8734: replayable canonical scope tension

```text
teamleaderleo/cloud-hypervisor@439ff52249e819f41570a5f3f3bf535d4bfb3e6e
base 1b004a7459ac752e1d7ad5a48237a1cb8608003b
pci/src/vfio.rs
```

Reading the actual patch confirmed the canonical case: the change adds `#[cfg(test)] mod unit_tests` immediately before an existing `#[cfg(test)] mod tests`.

The exact replay emitted one `test-module-precedent-tension` finding:

```text
changed declaration: unit_tests
repository precedent excluding changed declaration: unit_tests=88 of 127
same-file existing precedent: tests
repository-wide and file-local precedent disagree
change follows repository-wide precedent and differs from file-local precedent
repository evidence alone does not establish which scope should govern
```

The repository scan at the same head found 128 test-gated modules:

```text
unit_tests=89
tests=33
test_util=4
external_fds_tests=1
mock_vmm=1
```

An earlier draft of this PR incorrectly classified #8734 as a moved/lost fixture after looking only at title/head metadata. Reading the patch corrected that inference; the #16 comment and registry were repaired.

### Stensibly #1575: adapter gap

Exact case:

```text
head 78a9061b2feebe71211f0034ff2705b7143f6ce9
base 85cecf2608ad9e734a67518577fa85b9a08a550c
```

[...]

Issue #16, opened 2026-08-18
Create a dogfood corpus and precision-focused evaluation harness


teamleaderleo:
## Idea

Cultist should optimize for useful questions per developer interruption, not raw finding count.

Build a small corpus of real changes with known review outcomes and use it to evaluate new analyzers.

## Initial cases

- Cloud Hypervisor PR #8734: repository-wide `unit_tests` vs file-local `tests`
- Cultist's own PRs: self-diff should stay quiet when changes follow precedent
- future examples where a first exception is intentional
- future examples where a repeated review comment led to a real improvement

## What to measure

- did the relevant question appear?
- did the tool expose the right evidence/counterexamples?
- did it invent a rule that the repository does not have?
- number of low-value findings per change
- whether the finding changed after counterexample search
- runtime/indexing cost

Avoid reducing evaluation to one aggregate accuracy score. Keep fixtures readable enough that humans can inspect why a result is good or bad.

## CI / dogfood

Maintain a tiny self-dogfood step in normal CI. Larger external corpus tests can run separately or on demand so normal contributors are not blocked by network-heavy fixtures.

## Success criteria

- New analyzers can be tested against stable real-world cases.
- Regressions toward noisy/overconfident output are detectable.
- The Cloud Hypervisor case remains a canonical test for precedent tension.

teamleaderleo:
External corpus execution lane is now live via merged #129.

Durable pieces:

- `scripts/external_dogfood.py` runs the current Cultist binary against a read-only checkout and captures JSON + `CULTIST_PERF` receipts;
- `.github/workflows/external-dogfood.yml` is manual-only and read-only, with bounded checkout depth/history inputs;
- `research/external-repository-dogfood.md` records the execution policy and first real receipt.

Executed SmolRunner control at `teamleaderleo/smolrunner@ed3b70e375a57eabce26f2311f798f75b33bdeb0`, target `src/disposable_clone_runtime.rs`:

```text
cold scan: 4 Git, 259 parses,   0 hits, 591307 us, 4 findings
warm scan: 4 Git,   0 parses, 259 hits,  31640 us, 4 findings
ci-tests:  1 Git,   0 parses,   0 hits,   1197 us, 0 findings
history:   2 Git,   0 parses,            9842 us, 14/14 commits
```

Carrier run `32240366281`, artifact `9360596386`, digest `sha256:9794b9958627cb1b88d5f347496a1fc76b720d789ab8f66f1a48067989f2bf2b`.

[...]

PR #160, opened 2026-08-19
research: add bounded offline project-memory packet

teamleaderleo:
## Goal

Start #18 with the smallest useful project-memory primitive: a bounded offline packet for explicit PR/issue relationships, exact revision coordinates, selected evidence text, and changed paths.

Remote/provider access remains outside the Cultist process. This PR adds no product CLI command and no network behavior.

## Packet v1

`src/project_memory.rs` defines a strict JSON packet containing:

```text
repository owner/name
anchor artifact
bounded artifact list
bounded explicit edge list
```

Artifacts preserve:

- `pull_request | issue` identity;
- title/state/date coordinates;
- exact lowercase 40-hex PR head/base coordinates;
- changed paths for PRs;
- bounded retained evidence text;
- whether the retained text is complete.

Edges are typed:

```text
closes
follow_up_to
continuation_from
parent
related
```

Every edge carries the exact retained excerpt that establishes the relationship. Validation requires that excerpt to occur in the source artifact's admitted evidence text and requires both endpoints to exist in the packet.

Dates, vector order, nearby PR numbers, similar titles, and path overlap create no relationship edges.

## Fail-closed bounds

The parser rejects:

- input above 256 KiB;
- unknown schema versions/fields;
- malformed repository coordinates;
- duplicate or missing artifact identities;
- PRs without exact head/base coordinates;
- issues carrying PR revision/path fields;
- malformed repository-relative paths;
- missing/self edge targets;
- edge evidence absent from source artifact text;
- inconsistent open/closed timestamps.

## Retained Stensibly discriminator

`research/project-memory/stensibly-1575.json` pins five public provider artifacts:

```text
#1569  hosted-mail Convex runtime repair
#1571  semantic-admission index-length repair
#1573  disposition index-length repair
#1574  follow-up issue for the repeated identifier class
#1575  repository-wide Convex identifier-limit guard
```

Anchor #1575 explicitly says:

```text
Closes #1574.
Follow-up to production deployability repairs #1569/#1571/#1573.
```

The packet also retains exact continuation text from #1571 and #1573.

The semantic distinction stays visible:

- #1569 is a deployability predecessor clearing a Node-runtime bundling blocker;
- #1571 and #1573 are the two concrete overlong-index repairs;
- #1575 explicitly says ordinary CI accepted two overlong `by_*` identifiers and adds one repository-wide 64-character guard.

V1 therefore records enough explicit evidence to ask about lesson promotion without manufacturing a three-instance common failure class or inferring causality from timing.

## Research reader

```text
cargo run --example project_memory_packet < research/project-memory/stensibly-1575.json
```

The reader is bounded and offline and prints only the validated anchor summary + explicit anchor links.

## Tests

`tests/project_memory.rs` keeps the retained fixture in the standard matrix and asserts:

- #1575 exposes exactly three explicit follow-up PRs and one closing issue;
- five dated artifacts produce only the four anchor links actually backed by retained text;
- invented relationship evidence rejects;
- missing targets reject;
- issue/PR coordinate type confusion rejects;
- oversized input rejects before JSON parsing.

## Boundary / next step

This is an evidence packet, not a `why` finding or causal graph.

The next provider-side experiment can collect a selected anchor plus exact named artifacts into this offline form. Explicit references stay separate from chronological adjacency. Reviews/comments should enter only when a concrete replay earns that extra evidence type.

The Stensibly adapter-gap case from #138 now has an executable acceptance target for the first project-memory collector. Linux Fieldwork remains the complementary issue/investigation/note intake case.

Progresses #18
Progresses #41
Related: #16 #138

### DIFF
--- a/examples/project_memory_packet.rs
+++ b/examples/project_memory_packet.rs
@@ -2,7 +2,7 @@
 mod project_memory;
 
 use std::error::Error;
-use std::io::{self, Read};
+use std::io::{self, ErrorKind, Read};
 
 use project_memory::{MAX_PROJECT_MEMORY_BYTES, parse_project_memory_packet};
 
@@ -18,8 +18,11 @@ fn run() -> Result<(), Box<dyn Error>> {
     io::stdin()
         .take((MAX_PROJECT_MEMORY_BYTES + 1) as u64)
         .read_to_end(&mut input)?;
-    let packet = parse_project_memory_packet(&input)?;
-    let summary = packet.summary()?;
+    let packet = parse_project_memory_packet(&input)
+        .map_err(|error| io::Error::new(ErrorKind::InvalidData, error))?;
+    let summary = packet
+        .summary()
+        .map_err(|error| io::Error::new(ErrorKind::InvalidData, error))?;
     println!("{}", serde_json::to_string_pretty(&summary)?);
     Ok(())
 }
--- a/examples/project_memory_packet.rs
+++ b/examples/project_memory_packet.rs
@@ -1,0 +1,25 @@
+#[path = "../src/project_memory.rs"]
+mod project_memory;
+
+use std::error::Error;
+use std::io::{self, Read};
+
+use project_memory::{MAX_PROJECT_MEMORY_BYTES, parse_project_memory_packet};
+
+fn main() {
+    if let Err(error) = run() {
+        eprintln!("project-memory-packet: {error}");
+        std::process::exit(1);
+    }
+}
+
+fn run() -> Result<(), Box<dyn Error>> {
+    let mut input = Vec::new();
+    io::stdin()
+        .take((MAX_PROJECT_MEMORY_BYTES + 1) as u64)
+        .read_to_end(&mut input)?;
+    let packet = parse_project_memory_packet(&input)?;
+    let summary = packet.summary()?;
+    println!("{}", serde_json::to_string_pretty(&summary)?);
+    Ok(())
+}
--- a/research/project-memory.md
+++ b/research/project-memory.md
@@ -1,0 +1,167 @@
+# Project memory packet v0
+
+Issue #18 asks whether Git, PRs, issues, and reviews can become usable project memory while remote access remains optional.
+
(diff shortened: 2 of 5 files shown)


teamleaderleo (commented):
Skeptical pass found one semantic fail-closed blocker and one smaller contract mismatch.

### 1. Relation type is trusted independently of the retained excerpt

`ProjectMemoryPacket::validate()` proves only:

```text
edge.evidence occurs inside source.evidence_text
```

It never proves that `edge.relation` is the relation actually expressed by that excerpt.

So this malformed packet can currently pass conceptually:

```text
source evidence_text: "Related: #1574"
edge.relation: closes
edge.to: issue #1574
edge.evidence: "Related: #1574"
```

The packet summary then presents a typed `closes` link even though the admitted evidence says only `related`. The same confusion can happen among `parent`, `follow_up_to`, and `continuation_from`.

That is stronger than the source evidence and cuts against this PR's stated v1 invariant that an edge means "artifact A explicitly says relationship R to artifact B."

I would require a negative control where the target and excerpt are real but the relation kind is deliberately mismatched. V1 can fix this in either of two narrow ways:

- validate each admitted relation against an intentionally tiny exact phrase grammar; or
- make the packet distinguish a producer-asserted classification from the literal source excerpt, keeping the classification's admission/provenance explicit instead of presenting substring membership as validation of R.

Given the current five-relation vocabulary, the first option looks small enough if each grammar stays deliberately reviewed.

### 2. Timestamp validation is weaker than the documented contract

`validate_timestamp()` currently checks only non-empty/bounded/no-newline text. Values such as `banana` pass, and a closed artifact can have `closed_at` lexically/chronologically before `created_at`. The PR body says inconsistent open/closed timestamps reject and describes date coordinates as preserved evidence.

Either tighten v1 to an exact accepted timestamp form + ordering, or weaken the documented claim so consumers know timestamps are opaque supplied strings in this slice.

The rest of the boundary looks strong: exact PR revisions, issue/PR type separation, canonical paths, bounded evidence, explicit endpoints, and chronology/path overlap staying out of edge creation. The relation/excerpt mismatch is the one I would fix before treating the offline packet as typed project memory.

Issue #176, opened 2026-08-19
Make `check` help and parser errors identify the `check` command


teamleaderleo:
## UX gap

Merged #140 added `cargo cultist check` as a thin alias over the existing diff analyzer. Execution semantics are shared correctly, but the command still routes through `run_diff(args)`, `parse_diff_args(args)`, and `print_diff_help()`.

Current consequences:

```text
cargo cultist check --help
```

prints a `cargo-cultist diff` header/usage, and malformed `check` arguments report guidance such as:

```text
unknown diff option `...`; try `cargo cultist diff --help`
```

The analyzer alias is truthful; the UX currently names the implementation path instead of the command the caller used.

## Small repair

Preserve one analyzer/parser implementation while carrying the invoked command name into help/error rendering.

Conceptually:

```text
run()
  sees check | diff
  -> run_change(command_name, args)

parse_change_args(args, command_name)
  -> same DiffArgs
  -> errors say `cargo cultist {command_name} --help`

print_change_help(command_name)
  -> same semantic description
  -> header/usage use invoked command
```

`diff` remains fully supported and byte/semantic analyzer output stays identical for equivalent arguments.

## Controls

- `check --help` contains `cargo cultist check` and does not present itself as the diff command;
- `diff --help` continues to contain `cargo cultist diff`;
- unknown `check` option points to `cargo cultist check --help`;
- unknown `diff` option points to `cargo cultist diff --help`;
- parsing valid `check` and `diff` args yields identical `DiffArgs`;
- both commands still call the same `build_diff_analysis_report` path.

## Boundary

No analyzer, finding, report, or exit-code semantics change. This is command identity/help text only.

Refs #140.

teamleaderleo:
Completed by merged #177 (`c74b2c0c10bae1554408d7402eb62a178ee25ff1`). `check` and `diff` still parse to the same `DiffArgs` and enter the same `build_diff_analysis_report` path, while help/parser errors now identify the command the caller actually invoked.

PR #140, opened 2026-08-19
product: foreground decision-changing evidence

teamleaderleo:
## Goal

Turn the skeptical product review into a complementary product/evaluation lens while keeping the active research program moving.

Refs #137.

## Product wording

README/roadmap/agent guidance now ask an additional question alongside provenance and precision:

> Did the selected evidence change the next justified action, prevent an expensive wrong turn, or reduce important repository knowledge the next worker would otherwise rediscover?

Dogfood receipts can record outcomes such as:

```text
surfaced -> consulted -> changed next action
surfaced -> prevented/reversed a wrong turn
surfaced -> useful context, same action
surfaced -> ignored / irrelevant / stale / needed stronger evidence
candidate evidence stayed quiet -> correct negative
```

This is intentionally an inspectable behavioral lens rather than one opaque actionability/risk score.

The docs explicitly keep JEI, review, applicability, decision-memory, representation, analyzer, and performance research active in parallel.

## `check` alias

Add:

```text
cargo cultist check [--base REV] [--format text|json] [PATH]
```

`check` is a thin alias over the existing `diff` path:

- same argument parser;
- same `build_diff_analysis_report` analyzer;
- same text/JSON report model;
- `diff` remains supported.

This gives the during-work lifecycle a task-oriented verb without creating a second analyzer or changing report semantics.

## Collaboration dogfood

This lane started from then-current main. While it was being prepared, #126 and #132 merged. Before opening this PR I:

1. detected the branch was two commits behind;
2. inspected those landed changes;
3. confirmed they only added projection/evidence-role tests;
4. rebuilt this commit on the new exact main head;
5. rechecked current open PRs (#133, #136, #138, #139) and confirmed their changed-file sets are disjoint from this lane.

The PR is therefore one commit ahead / zero behind at open time.

## Validation

The ordinary repository CI should exercise rustfmt, Clippy, full tests, repository dogfood, history/CI-test analyzers, and PR diff text/JSON. The new unit control verifies `check` and `diff` route through the same change-command discriminator.

No analyzer semantics, finding schema, or authority behavior are changed.

PR #175, opened 2026-08-19
research: add descriptive behavioral episode summary

teamleaderleo:
## Goal

Give #137 an inspectable aggregate view over the unique raw episodes landed in #169 without inventing a universal actionability score.

## Summary v1

One already-validated `BehavioralEpisodeBatch` compiles to:

```text
total_episodes
surfaced
quiet
consulted
by_outcome[]
by_evidence_kind[]
```

Each grouped count retains:

```text
key
count
episode_ids[]
```

so every number stays traceable to the exact raw observations it summarizes.

No weighting exists. `changed_next_action=2` and `correct_quiet_negative=1` remain separate outcome counts rather than points in one score.

## Current retained corpus

The three #169 episodes summarize as:

```text
3 total
2 surfaced
1 quiet
2 consulted

2 changed_next_action
1 correct_quiet_negative
```

Evidence families remain distinct:

```text
active-work-heads-up
concurrent-main-head-movement
project-memory-relation-strengthening
```

The tests require every count's `count == episode_ids.len()` and deterministic sorted episode IDs.

## Research command

```text
cargo run --example behavioral_summary \
  < research/behavioral-episodes/cultist-collaboration-v1.json
```

Input is revalidated through the merged behavioral episode/receipt contracts before summarization.

## Boundary

- no actionability/risk/quality score;
- no analyzer promotion/demotion;
- no inference that an action change improved final task outcome;
- no cross-task normalization;
- no worker/model taxonomy;
- no timestamp/causality inference;
- raw episode IDs remain attached to every grouped count.

The useful next input is more varied raw episodes, especially currently empty outcomes such as ignored, irrelevant, stale/wrong-coordinate, needed-stronger-evidence, useful-same-action, and prevented/reversed-wrong-turn.

The branch is one commit ahead / zero behind current main `a551a948db945681195d5d644116090bb9ffc0db` at open time.

Refs #137 #157 #165 #169.

PR #177, opened 2026-08-19
fix: keep check command identity in help and errors

teamleaderleo:
## Problem

Merged #140 added `cargo cultist check` as a thin alias over the existing diff analyzer. Analyzer/report semantics are shared correctly, but the command still routed through diff-specific help/parser wording.

Examples before this patch:

```text
cargo cultist check --help
  -> header/usage presents `check / diff` together

cargo cultist check --bogus
  -> unknown diff option; try `cargo cultist diff --help`
```

Refs #176.

## Repair

Carry the invoked command name (`check` or `diff`) through the existing shared change path:

```text
run()
  -> run_diff(command, args)
  -> parse_diff_args(args, command)
  -> build_diff_analysis_report(...)   // unchanged
```

`DiffArgs`, analyzer selection, report construction, output formats, and exit behavior stay shared.

Help now renders one command-specific surface:

```text
cargo-cultist check
USAGE:
  cargo cultist check ...
```

or the equivalent `diff` form.

Parser errors point to the invoked command's own help.

## Controls

- valid `check` and `diff` arguments parse to identical `DiffArgs`;
- `check` help identifies check and does not present itself as a diff header;
- `diff` help identifies diff;
- unknown `check` option points to `cargo cultist check --help`;
- unknown `diff` option points to `cargo cultist diff --help`;
- both still enter the same `build_diff_analysis_report` path.

## Boundary

No analyzer, finding, report, or evidence semantics change. This is command identity/help text only.

The branch is one commit ahead / zero behind current main `df5ae59c757b56eefe023766429aca2c6ce9d800` at open time.

Refs #140 #176.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -86,8 +86,8 @@ fn run() -> Result<(), Box<dyn Error>> {
     }
 
     if args.first().is_some_and(|arg| is_change_command(arg)) {
-        args.remove(0);
-        return run_diff(args);
+        let command = args.remove(0);
+        return run_diff(&command, args);
     }
 
     if args.first().is_some_and(|arg| arg == "preflight") {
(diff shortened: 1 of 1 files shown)


teamleaderleo:
Rebuilt the one-file `check` command identity patch after its formatter-only failure. The final test now uses rustfmt's single-line assertion form. Main had advanced through #175/#178; this head is rebuilt directly on exact current main `c2133131038f33a98f7bc7d206ca6e4284be420a`. New head: `b28f71c8f7fe66b3cf2c3ca98e4a755452269f14`; command parsing/help semantics are unchanged from the prior head.

PR #178, opened 2026-08-19
research: retain semantic preflight collision episode

teamleaderleo:
## Goal

Grow the #137 behavioral corpus with a real **zero-path-overlap semantic collision** after merged #175 added descriptive counts.

## New retained episode

#166 and #167 changed disjoint files while tightening opposite sides of the same project-memory packet contract:

```text
#166 consumer
  strict single-line relation admission + exact target mention

#167 producer
  Primary case:\nURL issue evidence block
```

Both changes landed. Current-main integration then exposed that the #167 packet species no longer passed the #166 consumer boundary, and #174 repaired the producer/consumer seam.

The new receipt records:

```text
evidence_kind = project-memory-contract-collision
delivery = surfaced
consulted = true
outcome = changed_next_action
action = open #172 + implement #174
```

with episode ID:

```text
project-memory:primary-case-contract-collision:9792bfe->df5ae59
```

## Corpus growth

`cultist-collaboration-v1.json` grows from 3 to 4 unique episodes:

```text
4 total
3 surfaced
1 quiet
3 consulted

3 changed_next_action
1 correct_quiet_negative
```

Evidence families remain separate:

```text
active-work-heads-up
concurrent-main-head-movement
project-memory-contract-collision
project-memory-relation-strengthening
```

## Regression controls

- retained batch still passes merged #165 unique-episode validation;
- standalone #174 collision receipt is byte-for-semantics identical to its nested episode;
- merged #175 descriptive summary naturally updates from 3 -> 4 total and 2 -> 3 changed-next-action episodes;
- every summary count still carries its exact episode IDs.

## Why this is useful

This is a concrete corpus case for semantic preflight research under #96:

```text
Git/path overlap: none
shared contract: collided
integration review: changed next action
```

The receipt says what happened without claiming that Cultist can detect this species automatically yet.

## Boundary

- no actionability/risk/quality score;
- no analyzer promotion/demotion;
- no causal inference from disjoint paths alone;
- no claim that every producer/consumer change is related;
- raw episode identity + source coordinates remain inspectable.

The branch is one commit ahead / zero behind current main `75470b3fa146f50a57d1ac87413bc29985ed9b6e` at open time.

Refs #96 #137 #166 #167 #174 #175.

PR #174, opened 2026-08-19
fix: compose Primary case project-memory evidence

teamleaderleo:
## Live semantic collision

Merged #166 and #167 touched disjoint file sets while changing opposite sides of the same project-memory packet contract.

#166 tightened offline edge admission to one explicit relation line whose relation and `#N` target are independently validated.

#167's issue collector emits exact source evidence shaped as:

```text
Primary case:

https://github.com/OWNER/REPO/issues/N
```

with `relation=related`.

Its carrier passed before #166 landed. Current main after both merges rejects that packet species: the evidence is multiline, `Primary case:` is outside the ordinary line classifier, and the target appears as an issue URL rather than `#N`.

Refs #172.

## Repair

Keep #166's single-line grammar unchanged for ordinary relationship prose.

Multiline edge evidence now takes a separate fail-closed path which admits only:

```text
Primary case:
<optional blank lines>
https://{github.com|redirect.github.com}/PACKET_REPOSITORY/issues/N
```

The validator proves:

- exactly one non-empty label + one non-empty target URL;
- label is `Primary case:` case-insensitively;
- URL origin is admitted;
- URL repository equals packet repository;
- issue number is canonical positive decimal;
- typed relation is exactly `related`;
- parsed issue number equals the declared edge target.

Every other multiline shape rejects.

## Controls

New standard-suite tests require:

- exact #167-shaped block accepted;
- redirect.github.com form accepted;
- Primary case evidence cannot be strengthened to `closes`;
- cross-repository target rejects;
- URL/edge target mismatch rejects;
- extra non-empty prose rejects;
- multiline `Related: #N\n...` cannot bypass #166's single-line rule.

Existing ordinary project-memory packets continue through the unchanged single-line classifier.

## Dogfood result

This is a concrete positive for semantic preflight research:

```text
#166 paths != #167 paths
Git overlap: none
semantic packet contract: collided
integration review after both landed: caught it
```

Exact-path preflight was correctly unable to claim independence; this episode gives future semantic-preflight work a real producer/consumer contract collision with zero path overlap.

The branch is one commit ahead / zero behind current main `9792bfeeb08404f9b798be42df61d2b950c58e62` at open time.

Refs #96 #137 #160 #166 #167 #172.

PR #168, opened 2026-08-19
research: evaluate counterexample cohort refinements

teamleaderleo:
Progresses #142 with a research-only evaluator over supplied typed facts and support/counterexample observations.

## Result

The evaluator preserves baseline plus every discriminator partition and reports `candidate`, `no_improvement`, `overfit`, `unknown_current`, or `incomplete_evidence` without a scalar confidence score.

A `candidate` now requires all of these v0 conditions:

```text
counterexample discriminator evidence complete
excluded_counterexamples > 0
current support >= min_support
current counterexample proportion < baseline counterexample proportion
```

The proportion test uses exact integer cross multiplication over `u128`; no floating-point rounding or confidence score is introduced.

## Retained controls

- Oxc forward edit-class refinement: raw `99/100` -> syntax cohort `99/99` remains `candidate`;
- reverse Oxc `94/100` unchanged cohort remains `no_improvement`;
- singleton commit identity remains `overfit`;
- missing counterexample discriminator fact remains `incomplete_evidence`.

## #184 adversarial control

Issue #184 identified a false candidate:

```text
baseline: support=100, counterexamples=2, opportunities=102
selected: support=3,   counterexamples=1, opportunities=4
excluded: support=97, counterexamples=1
```

The previous implementation saw one excluded counterexample plus `min_support=3` and returned `candidate`, even though the counterexample rate regressed from `2/102` to `1/4`.

A dedicated regression test first proved the live failure on head `e9122a482eaec6f27e75cacd452a05faa1c0c20b`:

```text
CI run 32266326968 / job 96111610313 — failure
left:  Candidate
right: NoImprovement
```

The repaired head `85a9158d20bafbb3ec161d9283ca7f06b4ed5778` classifies that exact split as `no_improvement` using:

```text
current.counterexamples * baseline.opportunities
  < baseline.counterexamples * current.opportunities
```

## Validation

Exact repaired-head gates:

```text
CI run 32266580406 / job 96112440327 — success
Generated provenance review run 32266580474 — success
```

Full CI passed strict Clippy, all tests including the new ratio regression, project/review/closure/reference controls, active-work preflight, repository/history/CI-filter dogfood, and diff text + JSON dogfood.

Files: `src/cohort_refinement.rs`, `tests/cohort_refinement.rs`, `tests/cohort_refinement_ratio.rs`, `examples/cohort_refinement.rs`, `research/cohort-refinement.md`.

Research only; no product CLI/report-schema change.

Refs #6 #34 #35 #142 #184.

PR #166, opened 2026-08-19
fix: validate project-memory relation semantics

teamleaderleo:
## Problem

Merged #160 proves that an edge excerpt occurs in source `evidence_text`, but it does not prove the typed `relation` agrees with that excerpt.

So this can currently pass the offline boundary:

```text
evidence: "Related: #1574"
relation: closes
target: issue #1574
```

The packet then strengthens literal `related` evidence into a typed `closes` link.

Refs #163.

## Repair

For every edge, the Rust validator now requires all of:

1. edge evidence is one bounded single relationship line;
2. that exact line occurs in the source artifact evidence text;
3. the line belongs to the deliberately admitted relationship vocabulary;
4. the classified relation equals `edge.relation`;
5. the same line explicitly contains the declared target `#N` using canonical positive-decimal reference syntax.

Current classifier mirrors the narrow v1 vocabulary used by the landed #162 provider collector:

```text
Closes/Close/Closed/Fixes/Fix/Fixed/Resolves/Resolve/Resolved -> closes
Follow-up to / Follow up to                                -> follow_up_to
Continuation from / Deployment continuation               -> continuation_from
Parent:                                                     -> parent
Related:                                                    -> related
```

Unsupported prose stays unclassified.

## Adversarial controls

New standard-suite tests require rejection of:

- `Related: #1574` relabelled `closes`;
- `Parent: #1569` relabelled `follow_up_to`;
- a valid follow-up line reused with another existing target it does not mention;
- arbitrary `See #1574 for background.` prose given a typed relation.

The retained Stensibly #1575 packet still has to validate unchanged.

## Collaboration receipt

This repair started on merged #160 main. While it was being prepared, #162 merged the GitHub collector/workflow. I inspected that change, confirmed it is disjoint from `src/project_memory.rs`, rebuilt this fix on exact new main `66b13054f5c7e29d89acb580fe8e2465aaf1a779`, and retained the collector files.

The branch is one commit ahead / zero behind at PR open time.

## Boundary

- exact line grammar only; no fuzzy NLP;
- relation admission does not grant authority or infer causality;
- chronology/path overlap still create no edge;
- mixed-relationship prose on one line remains a producer-side precision concern from the #162 review;
- timestamp strictness from #163 remains a separate follow-up.

Refs #18 #160 #162 #163.

PR #179, opened 2026-08-19
research: preserve analyzer refinement episodes

teamleaderleo:
Progresses #148 with a bounded research ledger for counterexample-guided analyzer refinement.

V0 records supplied source receipts and replay outcomes rather than reimplementing domain analyzers. Each episode preserves the broad hypothesis, counterexamples, admitted discriminators, competing candidate refinements, rejected alternatives, one optional selected transition, source receipts, and optional links to already-observed behavioral episode IDs.

## Retained families

1. **#141 -> #144 justification/open obligation** — mandatory observed clearing-edge hypothesis -> durable open-before-evidence counterexample -> weakened zero-edge OPEN semantics.
2. **#142/#168 Oxc edit-class cohort** — raw 99/100 forward cohort -> supplied `edit_class` yields 99/99; reverse 94/100 remains `rejected_no_improvement`; singleton commit identity remains `rejected_overfit`.
3. **#166/#167 -> #174 project-memory contract** — zero-path-overlap producer/consumer collision -> split admission preserving strict ordinary relation grammar while admitting only exact same-repository `Primary case:` issue blocks.

The third family exercises the real behavioral join: its `behavioral_episode_ids[]` points at merged #178 episode `project-memory:primary-case-contract-collision:9792bfe->df5ae59`, and the standard test resolves that ID in the retained behavioral corpus and requires outcome `changed_next_action`.

## Meta contract

Validation owns bookkeeping only:

- bounded unique episode/candidate/discriminator identities;
- candidate discriminator refs must be admitted by the episode;
- selected transitions must name kept candidates;
- kept candidates retain expected replay cases, resolve a counterexample, and cannot have a failed held-out result;
- no-improvement/lost-case statuses must agree with supplied replay counts;
- rejected overfit remains a source-experiment conclusion;
- behavioral links are explicit and never fabricated.

Domain replay truth stays in the exact source receipts.

The original five semantic blobs were transplanted byte-for-byte onto current `main@51affea6c645780652e6af9c80d79e4763eb609b`, erasing stale branch ancestry without changing the ledger:

```text
head: f0accb7832c6f8509affe3d64e5789d1d11265cb
commits: 1
files: 5
```

Exact-head ordinary CI:

```text
run: 32268251362
job: 96117991420
result: success
```

That run passed formatter, strict Clippy, project-memory lineage controls, GitHub review-memory and issue-closure adapter controls, external GitHub reference guards, active-work preflight, full tests, and repository/history/CI-filter/diff dogfood.

Generated-provenance review also passed:

```text
run: 32268251406
job: 96117992037
result: success
```

`main` remained at the exact reanchor base through both gates.

The earlier executed research receipt remains in `research/refinement-episodes.md`; this current-main run proves the same bytes compose with the modern repository surface.

Files: `src/refinement_episode.rs`, `tests/refinement_episode.rs`, `examples/refinement_episodes.rs`, `research/refinement-episodes/cultist-v1.json`, `research/refinement-episodes.md`.

## Boundary / next discriminator

Research only; no analyzer predicate language, scalar score, automatic promotion, or product CLI/report-schema change.

Three families expose three different discriminator-production mechanisms. The next earned child is the source-owned discriminator observation layer in #187.

Refs #137 #141 #142 #144 #148 #166 #167 #168 #174 #175 #178 #187.

### DIFF
--- a/examples/refinement_episodes.rs
+++ b/examples/refinement_episodes.rs
@@ -1,0 +1,33 @@
+#![allow(dead_code)]
+
+use std::error::Error;
+use std::io::{self, Read};
+
+#[path = "../src/refinement_episode.rs"]
+mod refinement_episode;
+
+use refinement_episode::{MAX_REFINEMENT_EPISODE_BATCH_BYTES, parse_refinement_episode_batch};
+
+fn main() {
+    if let Err(error) = run() {
+        eprintln!("refinement-episodes: {error}");
+        std::process::exit(1);
+    }
+}
+
+fn run() -> Result<(), Box<dyn Error>> {
+    let mut bytes = Vec::new();
+    io::stdin()
+        .take((MAX_REFINEMENT_EPISODE_BATCH_BYTES + 1) as u64)
+        .read_to_end(&mut bytes)?;
+    if bytes.len() > MAX_REFINEMENT_EPISODE_BATCH_BYTES {
+        return Err(format!(
+            "refinement episode batch exceeds the {MAX_REFINEMENT_EPISODE_BATCH_BYTES}-byte limit"
+        )
+        .into());
+    }
+
+    let batch = parse_refinement_episode_batch(&bytes)?;
+    println!("{}", serde_json::to_string_pretty(&batch)?);
+    Ok(())
+}
--- a/research/refinement-episodes.md
+++ b/research/refinement-episodes.md
@@ -1,0 +1,260 @@
+# Counterexample-guided analyzer refinement episodes
+
+Tracking: #148. Built after #141/#144, #142/#168, project-memory contract refinement #174, and the behavioral episode work in #137/#175/#178.
+
+## Question
+
+Can Cultist preserve analyzer-research transitions as bounded, replayable records while leaving domain facts and predicates with the analyzers that produced them?
+
+V0 answers only the bookkeeping layer.
+
+## Episode contract
+
+```text
+RefinementEpisode
+  id
+  family
+  hypothesis_before
+  counterexample_refs[]
+  admitted_discriminators[]
+  candidate_refinements[]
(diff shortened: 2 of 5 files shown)


teamleaderleo (commented):
One meta-level referential-integrity gap: `behavioral_episode_ids` are documented as links to **already-observed #165 episodes**, but the validator currently treats them as bounded unique strings only. The test makes that explicit by admitting `observed:episode-1` without supplying or resolving any behavioral episode corpus.

That means a future refinement episode can carry:

```text
behavioral_episode_ids = ["made-up-action-changing-episode"]
```

and the ledger will serialize it indistinguishably from a real observed receiver episode. This is precisely the sort of join where bookkeeping semantics *do* own a boundary even though domain replay truth stays external.

I would keep the standalone parser lightweight and add a separate join validator, for example:

```text
validate_behavioral_links(refinement_batch, behavioral_episode_batch)
```

which:

- revalidates the #165 batch first;
- requires every non-empty `behavioral_episode_id` to resolve exactly once;
- rejects missing IDs;
- keeps the base refinement ledger usable without a behavioral corpus, but then describes those strings as unresolved/supplied refs until join validation occurs.

A retained control can use the current four-episode `research/behavioral-episodes/cultist-collaboration-v1.json`: one real ID should join, `made-up-action-changing-episode` should fail. That keeps the good separation in this PR—domain analyzer truth remains in source receipts—while preventing the ledger from manufacturing observed behavioral evidence by typo or optimistic annotation.

The selected/rejected candidate bookkeeping itself looks coherent from this pass: kept candidates cannot lose expected cases or resolve zero counterexamples, rejected-no-improvement preserves baseline replay, discriminators must be admitted, and selected transitions must name kept candidates.

Issue #180, opened 2026-08-19
Pilot JEI budgeting with positive and negative historical controls


teamleaderleo:
## Goal

Pilot the current JEI packet and semantic byte budget against **real historical repository episodes**, with both a positive case inside today's file-local capability and a negative case that exposes its scope boundary.

No model grader is required.

## Positive control: SmolRunner same-target earned history

Pinned state:

```text
repo   teamleaderleo/smolrunner
ref    ed3b70e375a57eabce26f2311f798f75b33bdeb0
file   src/disposable_clone_runtime.rs
```

Cultist #62 already recorded four earned-history discriminators for this exact target:

```text
Remove duplicate worker readiness probes (#540)
Reduce live admission latency before JIT
fix: collapse redundant clone admission polls (#533)
keep clone preflight before durable checkpoint (#530)
```

Acceptance:

- a 1 MiB control must recover all four;
- the default 32 KiB semantic budget must recover all four;
- smaller budgets are observational and record the first tested loss.

## Negative control: Stensibly distributed failure family

Pinned state:

```text
repo   teamleaderleo/stensibly
ref    85cecf2608ad9e734a67518577fa85b9a08a550c
file   convex/schema.ts
```

At this exact repository revision two same-class production deployability repairs already exist:

```text
#1571 Keep Gmail semantic-admission indexes deployable
#1573 Keep Gmail mailbox-disposition indexes deployable
```

They live in different sibling schema modules:

```text
convex/mailSemanticAdmissionSchema.ts
convex/gmailMailboxDispositionSchema.ts
```

Neither touched `convex/schema.ts`.

PR #1575 later generalized those repeated Convex identifier-limit failures into a repository-wide `convex/**/*.ts` guard. Its guard commit remains outside the pinned packet state.

Acceptance:

- the workflow must prove #1571 and #1573 exist in the sibling-file histories at the pinned revision;
- the 1 MiB file-local `convex/schema.ts` packet must miss both;
- the 32 KiB packet must also miss both;
- this is expected negative evidence that file-local JEI cannot recover a failure family distributed across sibling targets.

## Budget sweep

Both controls run the real schema-v5 `agent_context_packet` at:

```text
1 MiB
32 KiB
16 KiB
8 KiB
4 KiB
```

Receipts record:

- exact repository/revision/target;
- candidate bytes and selected bytes;
- semantic eviction receipts;
- per-lesson presence;
- whether the declared positive/negative expectation held.

## First executed result

[...]

PR #181, opened 2026-08-19
research: pilot JEI budgets with historical controls

teamleaderleo:
Starts #180 and advances #67/#106 with an executed positive/negative historical pilot matrix.

## Why two controls

The first Stensibly replay immediately proved that the expected #1571/#1573 lessons were absent even from a 1 MiB packet with zero semantic evictions. Investigation showed those repairs live in different sibling schema modules and neither touched the file-local target `convex/schema.ts`.

That turns the case into a useful **negative scope control**, rather than a failed budget benchmark.

The PR therefore pairs it with the already-earned **positive same-target control** from #62.

## Positive · SmolRunner same-target history

Pinned state:

```text
repo   teamleaderleo/smolrunner
ref    ed3b70e375a57eabce26f2311f798f75b33bdeb0
file   src/disposable_clone_runtime.rs
```

Required earned-history subjects:

```text
Remove duplicate worker readiness probes (#540)
Reduce live admission latency before JIT
fix: collapse redundant clone admission polls (#533)
keep clone preflight before durable checkpoint (#530)
```

Expectation: all four survive the 1 MiB control and 32 KiB acceptance budget.

Executed result:

- candidate packet ≈ 13.46 KiB;
- all four survived every tested budget: 1 MiB, 32 KiB, 16 KiB, 8 KiB, 4 KiB;
- 8 KiB selected 8014 bytes after evicting 16 support examples + 11 counterexamples;
- 4 KiB selected 4094 bytes after evicting 16 support examples, 16 counterexamples, 8 companion relations, and 5 older history summaries;
- minimum tested preserving budget: 4096 bytes.

## Negative · Stensibly cross-file failure family

Pinned state:

```text
repo   teamleaderleo/stensibly
ref    85cecf2608ad9e734a67518577fa85b9a08a550c
file   convex/schema.ts
```

The workflow separately proves the same repository snapshot contains:

```text
#1571 in convex/mailSemanticAdmissionSchema.ts history
#1573 in convex/gmailMailboxDispositionSchema.ts history
```

Neither appears in `convex/schema.ts` history.

Expectation: the file-local packet misses both at 1 MiB and 32 KiB.

Executed result:

- candidate packet ≈ 14.09 KiB;
- zero #1571/#1573 lessons at 1 MiB, 32 KiB, 16 KiB, 8 KiB, or 4 KiB;
- 1 MiB and 32 KiB required zero semantic evictions, proving the miss precedes budgeting;
- sibling-file history at the exact pinned revision contains both repair subjects;
- PR #1575 later generalized the repeated Convex identifier-limit failures into a repository-wide guard, but that later guard is never fed into the pre-guard packet.

## What this PR adds

- replayable `stensibly-1575-pre-guard-jei` corpus case while preserving the broader #1575 guard-promotion `adapter_gap` case;
- `scripts/jei_history_pilot.py` with explicit `preserve_all` and `miss_all` expectations;
- a reusable `JEI historical budget pilot` workflow;
- machine-readable per-control receipts and a combined job summary.

## Main result

```text
same-target earned history
  -> survives aggressive semantic budgeting

distributed sibling-file failure family
  -> file-local packet misses it before budgeting begins
```

For these first real controls, **scope selection is the stronger current limitation than packet size**.

The next discriminator should research bounded task/scope expansion beyond one file: explicit target sets, directory/package scope, relation-led expansion, or project-memory evidence. Any expansion still needs omission receipts and a hard semantic budget.

## Boundary

- read-only public external checkouts;
- no provider mutation;
- no external builds/deploys;
- no LLM/model grader;
- no primary Cultist CLI change;
- no claim that commit subjects alone prove causal equivalence;
- no weakening of C1 transport bounds or the broader Stensibly adapter gap.

Refs #41 #62 #67 #74 #106 #139 #180.

### DIFF
--- a/.github/workflows/jei-history-pilot.yml
+++ b/.github/workflows/jei-history-pilot.yml
@@ -1,0 +1,132 @@
+name: JEI historical budget pilot
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - ".github/workflows/jei-history-pilot.yml"
+      - "scripts/jei_history_pilot.py"
+      - "research/external-dogfood-cases.json"
+
+permissions:
+  contents: read
+
+jobs:
+  stensibly-1575-pre-guard:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Cultist
+        uses: actions/checkout@v4
+        with:
+          path: cultist
+
+      - name: Resolve pinned pilot case
+        id: target
+        run: |
+          python cultist/scripts/external_dogfood_case.py \
+            --registry cultist/research/external-dogfood-cases.json \
+            --case stensibly-1575-pre-guard-jei \
+            --github-output "$GITHUB_OUTPUT" \
+            > resolved-target.json
+          cat resolved-target.json
+
+      - name: Install stable Rust
+        run: |
+          rustup toolchain install stable --profile minimal
+          rustup default stable
+
+      - name: Build agent context packet research example
+        run: cargo build --release --example agent_context_packet --manifest-path cultist/Cargo.toml
+
+      - name: Checkout pinned Stensibly state
+        uses: actions/checkout@v4
+        with:
+          repository: ${{ steps.target.outputs.repository }}
+          ref: ${{ steps.target.outputs.ref }}
+          fetch-depth: ${{ steps.target.outputs.fetch_depth }}
+          persist-credentials: false
+          path: target
+
+      - name: Verify exact historical coordinate
+        env:
+          EXPECTED_HEAD: ${{ steps.target.outputs.ref }}
+        run: |
+          set -euo pipefail
+          actual="$(git -C target rev-parse HEAD)"
+          test "$actual" = "$EXPECTED_HEAD"
+          test -f "target/${{ steps.target.outputs.history_file }}"
+          git -C target log --no-merges --format='%H %s' -n 20 -- "${{ steps.target.outputs.history_file }}"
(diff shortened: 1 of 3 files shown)


teamleaderleo:
## First executed pilot result: useful negative control

Run `32247583166` reached the exact pinned Stensibly coordinate `85cecf2608ad9e734a67518577fa85b9a08a550c`, built the real schema-v5 `agent_context_packet`, and swept the declared budgets.

Result: the 1 MiB control already missed both expected #1571/#1573 lessons, with a ~14.1 KiB packet and zero semantic evictions. Therefore the 32 KiB failure is **not** a byte-budget failure.

The reason is now explicit from the source PRs:

- #1571 changed `mailSemanticAdmission*` files;
- #1573 changed `gmailMailboxDisposition*` files;
- neither changed the pilot target `convex/schema.ts`.

So the current single-file packet cannot reconstruct this repeated failure family from that unrelated file-local history. This is an earned scope-boundary result, not a reason to weaken the acceptance condition.

I am preserving this as the negative control and adding the existing #62 SmolRunner same-target replay as the positive control. The pilot will pass when:

```text
SmolRunner same-target history
  -> earned lessons survive 32 KiB

Stensibly distributed sibling-file failure family
  -> remains absent even in the 1 MiB file-local control
```

That separates semantic budgeting from scope selection and gives #106 a concrete next research target: task/scope-conditioned history beyond one file.

Issue #186, opened 2026-08-19
Research explicit scope history after JEI pilot


teamleaderleo:
## Trigger

The executed JEI pilot in #180/#181 produced a clean discriminator:

```text
SmolRunner same-target history
  -> all four earned lessons survive down to the 4 KiB test

Stensibly #1575 pre-guard
  -> file-local convex/schema.ts packet misses #1571/#1573 even at 1 MiB
  -> both repairs exist in sibling schema-file history at the same pinned revision
```

So the next question was selection scope, not a looser byte budget.

## Research question

> Can one explicitly requested, bounded repository-relative scope recover distributed sibling-file history while preserving the file-local packet as the default?

Exact Stensibly control:

```text
repo   teamleaderleo/stensibly
ref    85cecf2608ad9e734a67518577fa85b9a08a550c
target convex/schema.ts
scope  convex/
```

Required distributed lessons:

```text
Keep Gmail semantic-admission indexes deployable (#1571)
Keep Gmail mailbox-disposition indexes deployable (#1573)
```

The later #1575 generalized guard remains outside the checkout.

## First discriminator: external composition before packet-schema change

#189 composes:

```text
existing file-local agent_context_packet
+
explicit bounded scope_recent_history
```

without changing the Rust packet schema.

The scope component:

- accepts one explicit canonical repository-relative existing directory;
- requires the scope to contain the target and remain inside the same repository;
- runs bounded non-merge Git history only under that scope;
- deduplicates commits already present in target-local recent history;
- preserves exact commit SHA/date/subject;
- records scope-history truncation;
- carries explicit UNKNOWN that chronology alone does not prove semantic relevance;
- measures scope-history bytes and full combined-envelope bytes;
- rejects an over-budget combined envelope instead of byte truncation.

## Executed result

The discriminator passes.

At the exact pinned pre-guard revision:

- direct `git log -- convex` contains #1573 and #1571 as the two newest scope commits;
- the ordinary file-local `convex/schema.ts` packet still lacks them, preserving #181's negative control;
- the explicit `convex/` scope receipt recovers both.

### 1 MiB control

```text
file packet        14,087 bytes
scope history       2,687 bytes
combined envelope  19,148 bytes
lessons             both present
fits                yes
```

### 32 KiB acceptance

```text
file packet        14,085 bytes
scope history       2,687 bytes
combined envelope  19,144 bytes
lessons             both present
fits                yes
```

[...]

PR #191, opened 2026-08-19
feature: add explicit scoped agent context envelope

teamleaderleo:
Promotes the explicit-scope discriminator earned in #186/#189 into Rust while preserving the existing file-local research packet unchanged.

## Compatibility boundary

Existing behavior remains available exactly through:

```text
cargo run --example agent_context_packet -- FILE
```

The opt-in scoped surface is a separate research example:

```text
cargo run --example scoped_agent_context_packet -- FILE --scope DIR
```

That avoids silently changing the schema-v5 file packet for existing consumers.

## Scoped envelope

The Rust scoped example includes the existing `agent_context_packet` implementation internally and adds:

- one explicit canonical repository-relative directory scope;
- scope must exist inside the same Git repository and contain the target;
- bounded 20-commit non-merge scope history;
- SHA-based deduplication against target-local recent history;
- exact SHA/date/subject receipts;
- separate `scope_recent_history` and truncation signal;
- explicit UNKNOWN that scope chronology alone does not prove semantic relevance;
- a combined semantic byte budget.

## Budget priority

The combined compiler preserves the priority earned by #181/#189:

```text
scope recent-history summaries
  -> evict first

target-local packet evidence
  -> only then use the existing schema-v5 eviction ladder
```

Every scope-history eviction is receipted as `scope_recent_history_summary`. Existing file-packet semantic eviction receipts remain on the nested packet.

If scope history and all ordinary evictable file-packet detail are exhausted while the protected combined core still exceeds the budget, the scoped compiler fails closed.

## Standard-suite controls

The ordinary test suite now checks:

- traversal/absolute/noncanonical scope text rejects;
- scope-history SHA deduplication;
- scope history is evicted before target-local recent history under pressure.

The include-based research wrapper keeps helper code under `examples/support/` so Cargo does not compile support as an independent example target.

## Executed Stensibly promotion replay

The existing #189 Python reference and this Rust implementation run side by side against the exact pre-guard state:

```text
repo   teamleaderleo/stensibly
ref    85cecf2608ad9e734a67518577fa85b9a08a550c
target convex/schema.ts
scope  convex
budget 32768
```

The unchanged target-local history still excludes both distributed repairs. The explicit scope history recovers:

```text
Keep Gmail mailbox-disposition indexes deployable (#1573)
Keep Gmail semantic-admission indexes deployable (#1571)
```

### Rust result

```text
candidate bytes          18,794
selected bytes           18,794
scope history rows       18
scope history truncated  true
scope evictions          0
file-packet evictions    0
both lessons recovered   yes
```

The independent Python reference remains green at 19,144 bytes for the same 32 KiB case.

Validation also confirms:

- `--scope ../` rejects as noncanonical;
- `--scope src` rejects because it does not contain `convex/schema.ts`.

Workflow artifact:

```text
id      9363982985
sha256  4e689937d404f5b2d59820a71bf21eb690409faa9bb8d0c28e5348c0c0105726
```

Normal CI, the scoped unit/integration tests, and the historical Rust replay are green on the tested head.

## Result

The pilot progression is now executable end to end:

```text
file-local Stensibly packet
  -> misses distributed #1571/#1573 lesson

explicit convex/ scope discriminator
  -> recovers both

Rust scoped envelope
  -> reproduces recovery at 18.8 KiB under the 32 KiB semantic cap
```

That earns explicit requested scope history as a reusable Rust research capability. Automatic scope inference remains future research.

## Boundary

- explicit scope only;
- no automatic scope inference;
- no whole-repository default;
- no semantic-equivalence claim from chronology;
- no primary Cultist CLI change;
- no external mutation.

Refs #62 #67 #74 #106 #139 #180 #181 #186 #189.


teamleaderleo:
Budget-policy adversarial control opened as #196. The current compiler explicitly pops `scope_recent_history` before calling the file-packet eviction ladder. That is truthful, but #189 is a counterexample candidate because the two task-changing lessons exist only in scope history while file-local history missed them at 1 MiB. Please add a tight-budget control before promotion: preserve harmless target-local history, evict one of two modeled decision-changing scope lessons, and show whether the modeled next action changes. Then sweep the real Stensibly carrier below the ~19.1 KiB candidate size. If #1571/#1573 disappear before protected-core failure, target-local-first is a heuristic—not a JEI invariant.

teamleaderleo:
#197 negative control is green on the real scoped compiler: exact head `e8ffa78dd4bd3893f07a351e41846ee9766b3e8d`, CI `32249877846`. A small budget squeeze evicts scope history first while target-local recent history survives, and that changes the fixture-local next action. So the current `scope -> evict first` ordering is now explicitly a heuristic, not a decision-preservation invariant. Your real Stensibly carrier also establishes the current Rust candidate at 18,794 bytes with 18 scope-history rows and zero evictions at 32 KiB. #196 now wants the external sweep just below that boundary.

teamleaderleo:
The #196/#197 external Stensibly pressure sweep found the exact budget cliff on your current Rust scoped compiler. Run `32250098796`, artifact `9364078036`.

At 16,000 bytes the selected envelope is 15,956 bytes with exactly 2 scope-history rows left—#1573 and #1571—and all 17 target-local recent-history rows still present. At 15,750 bytes the compiler removes both final scope rows, emits 15,575 bytes, and still preserves all 17 target-local recent-history rows with zero file-packet evictions. Both demonstrated task lessons are gone.

The compiler continues to fit all the way to tested 4,096 bytes; protected-core failure never fires. So the 32 KiB acceptance remains valid, but `scope history evicts first` is now externally falsified as a decision-preservation invariant. A next policy experiment should accept explicit selected/protected evidence refs from the JEI layer and fail closed if those selected refs cannot coexist with the protected core—without inferring protection from recency/path locality.

Issue #190, opened 2026-08-19
Research missing discriminator observation frontiers into probe planning


teamleaderleo:
## Thesis

#179/#187 now establish two separate generic objects:

```text
selected refinement
  -> discriminator_refs[]

source-owned discriminator observations
  -> KNOWN | UNKNOWN | INVALID
```

#145/#164 establish a different source-owned execution vocabulary:

```text
DurableObligation.missing_discriminator
  DiscriminatorKey { kind, target }

EvidenceProbe
  produces DiscriminatorKey
  exact clearing requirements
  effect class
  forecast cost
```

The remaining bridge is:

> When a selected refinement needs discriminator D and no current KNOWN observation D exists, how does Cultist expose that missing observation and map it to an admitted probe without pretending the two vocabularies are identical?

## Phase A: missing-observation frontier

Given:

```text
required discriminator refs
+ #185 observation enumeration
```

produce a bounded receipt such as:

```text
ObservationFrontier
  discriminator_id
  subject_ref
  state
    missing
    unknown { observation_refs[] }
    invalid { observation_refs[] }
    current { observation_refs[] }
```

Rules:

- `current` requires at least one KNOWN observation for the required discriminator/subject;
- UNKNOWN remains UNKNOWN;
- INVALID remains INVALID/stale;
- zero observations is explicit `missing`;
- an observation for another subject cannot satisfy the frontier merely because the discriminator ID matches;
- multiple current observations remain inspectable rather than collapsing into one hidden truth value.

The frontier performs no source acquisition.

## Phase B: source-owned probe bridge

A separate adapter supplies an explicit mapping from one observation requirement to one #145 clearing/probe contract, conceptually:

```text
ObservationProbeBridge
  observation_discriminator_id
  observation_subject_ref
  probe_discriminator: { kind, target }
  clearing_requirements
  source_receipt
```

The bridge is evidence that a probe capable of producing the #145 discriminator is also an admitted producer for the missing #185 observation.

No mapping is inferred from string equality.

## First controls

1. **Mapped capable probe**
   - selected refinement requires D;
   - #185 has no current D;
   - exact source bridge maps D/S to one #145 `{kind,target}` + clearing coordinate;
   - #145 selects a capable admitted probe under its ordinary effect/cost policy.

2. **Similarly named unmapped probe**
   - probe `kind` resembles discriminator ID D;
   - no explicit bridge exists;
   - frontier remains unresolved for acquisition.

3. **Wrong subject**
[...]

PR #187, opened 2026-08-19
research: add source-owned discriminator observation refs

teamleaderleo:
Starts #185 as the first landed child of #179.

Adds a research-only typed reference seam for discriminator observations already produced by source analyzers. One observation carries bounded identity, discriminator ID, subject ref, source receipt, KNOWN/UNKNOWN/INVALID value state, and an optional applicability receipt reference.

Generic enumeration groups only KNOWN observations by discriminator/value reference while preserving every observation/source identity. UNKNOWN and INVALID observations remain explicit and never enter current partitions.

## Three-family composition

The retained corpus supplies current observations for every discriminator used by #179's selected transitions:

- `clearing_evidence_presence` from #159;
- `edit_class` from the retained Oxc syntax-cohort replay;
- `primary_case_evidence_form` and `same_repository_issue_target` from #174.

A cross-object test parses the now-landed #179 refinement corpus and requires every selected candidate discriminator ref to have at least one current KNOWN observation, without executing any source analyzer.

## Controls

- KNOWN values enumerate deterministically;
- UNKNOWN/INVALID stay visible and out of current partitions;
- duplicate and conflicting observation identities reject separately;
- missing source receipt rejects;
- equal values from distinct source receipts preserve both observation identities;
- JSON round trip/order and byte bounds are explicit;
- value spelling remains opaque and grants zero evidence strength/authority/disposition.

The module deliberately cannot decide overfit, replay improvement, applicability meaning, probe capability, or action.

## Current-main promotion receipt

The five #187 semantic blobs were rebuilt as one commit after #179 landed:

```text
head: 5e98b403229e76a7b0e7a5ce46cf75c5f0f8b60d
commits: 1
files: 5
```

The PR was retargeted to `main`. Its successful merge-view gate included current `main@0eabd231a23dfc243c167d336b7ad9836aff2bcc`, including the newer frozen first-action trial work.

Ordinary CI:

```text
run: 32268881806
job: 96120117040
result: success
```

Generated-provenance review:

```text
run: 32268881627
job: 96120116433
result: success
```

The full current CI surface passed: formatter, strict Clippy, project-memory lineage, GitHub review-memory and issue-closure controls, external-reference guards, active-work preflight, full tests, and repository/history/CI-filter/diff dogfood.

The earlier semantic receipt remains in `research/discriminator-observations.md`; this run proves the same observation layer composes with the current repository.

[...]

Issue #185, opened 2026-08-19
Research source-owned discriminator references for analyzer refinement


teamleaderleo:
## Thesis

#179/#148 now preserves three counterexample-guided refinement families in one generic ledger while keeping domain replay with the source analyzers:

```text
justification
  clearing-evidence presence

historical companion
  Rust edit class

project memory
  Primary case evidence form + repository/target contract
```

The next question is whether these families can expose a **small source-owned discriminator reference interface** that lets a refinement harness enumerate supplied candidate partitions without importing domain execution logic.

The goal is a reference seam, not a universal fact ontology.

## Candidate contract

Research the smallest object resembling:

```text
DiscriminatorObservation
  id
  source_receipt
  subject_ref
  value_state
    known(value_ref)
    unknown(reason_ref)
    invalid(reason_ref)
  applicability_ref?
```

Exact field names are provisional.

Important split:

```text
discriminator id
  names the distinction

source receipt
  identifies the evidence/evaluator that produced it

value/value_ref
  identifies the supplied current partition value

applicability
  says whether that observation may be used at the current coordinate
```

The refinement layer consumes these references. It does not calculate their domain meaning.

### A. Justification / durable UNKNOWN

Source family: #141/#144/#156/#159.

Candidate observation:

```text
id = clearing_evidence_presence
value = absent | observed
source = exact justification/durable-obligation receipt
```

Control: revision movement must continue to be decided by the owning applicability evaluator. The reference layer cannot reinterpret a stale receipt as current.

### B. Historical companion cohort

Source family: #35/#142/#168.

Candidate observation:

```text
id = edit_class
value = syntax_changed | comments_or_docs_only | unclassified
source = exact syntax-cohort/classification receipt
```

Control: exact commit identity remains available as a supplied discriminator but #168's overfit result remains source-owned. The generic reference type cannot decide that a partition is overfit merely from its spelling.

### C. Project-memory packet contract

Source family: #166/#167/#174/#178/#179.

Candidate observations:

```text
id = evidence_form
value = ordinary_relation_line | primary_case_issue_block

id = target_repository_relation
value = same_repository_issue | other
```

Control: parsing the `Primary case:` block and proving repository/target equality remain in project-memory validation. The reference object receives the result only.

## Key experiment

Build one bounded batch containing one or more supplied observations from each family and let a tiny generic reader validate only:

- bounded unique IDs;
- explicit source receipts;
- explicit subject identity;
- known/unknown/invalid state;
[...]

Issue #148, opened 2026-08-19
Formalize analyzer research as counterexample-guided refinement


teamleaderleo:
## Thesis

Cultist's existing research discipline already follows a powerful loop:

```text
hypothesis
-> deterministic probe
-> real discriminator
-> counterexample / negative control
-> keep, weaken, split, reject, or promote
```

Research whether this can become a reusable analyzer-development harness so candidate detectors are refined against explicit counterexamples instead of growing ad hoc special cases.

The aim is a deterministic detector foundry, not automatic policy synthesis.

## Candidate experiment contract

```text
ResearchHypothesis
  id
  admitted fact/discriminator vocabulary
  candidate predicate
  positive fixtures
  negative fixtures
  held-out fixtures
  current status

ReplayResult
  matched positives
  false positives
  missed positives
  counterexamples
  refinement receipt
```

## Refinement loop

1. evaluate candidate predicate on all visible fixtures;
2. retain exact failing fixture/counterexample;
3. propose or enumerate one refinement from an admitted deterministic discriminator vocabulary;
4. rerun positives + negatives;
5. reject refinements that merely memorize a fixture or erase intended positives;
6. keep competing refinements explicit;
7. stop with `retain research`, `weaken`, `split`, `reject`, or `candidate for promotion`.

A model may later suggest refinements, but deterministic replay decides what survived.

## First research targets

Use two existing Cultist families with rich counterexamples:

- historical/generated companion evidence;
- explicit coordination phrase extraction or precedent tension.

Require one case where a tempting broad hypothesis is weakened or split after a counterexample.

## Example

```text
H0
  source A change => generated B expected

counterexample
  docs-only A change correctly omits B

candidate H1
  code-bearing A change + proven generator ownership => B expected

replay
  positives retained
  docs-only negative removed
  held-out generated migration remains counterexample

result
  retain H1 as narrower research hypothesis
```

## Anti-overfit controls

- no exact commit SHA/path literals as a refinement unless the research question is explicitly path-scoped;
- minimum positive support where the family is empirical;
- held-out fixture check before promotion;
- original counterexamples remain linked to the refined hypothesis;
- chronology alone cannot become a causal discriminator;
- every discriminator must identify its producing evidence source.

## Composition

- #6 counterexample-first findings;
- #16 evaluation corpus;
- #29 Fieldwork research feeder;
- #41 agentic-history corpus;
- #142 cohort refinement;
- #11 lint incubation/promotion;
[...]

PR #194, opened 2026-08-19
research: expose missing discriminator observation frontiers

teamleaderleo:
Starts #190 Phase A as the exact-frontier child of the now-landed #187 observation layer.

Adds a read-only frontier over exact observation requirements `(discriminator_id, subject_ref)` and the source-owned observation batch.

## Frontier semantics

States are `current`, `unknown`, `invalid`, and `missing`. Current usability follows:

```text
current -> unknown -> invalid -> missing
```

while all matching receipts remain inspectable. Same-discriminator observations for another subject are retained under `other_subject` and never satisfy the exact requirement.

The standard harness derives the four selected #179/#187 requirements and verifies they are current, then covers removal -> missing, wrong subject -> missing, UNKNOWN/INVALID states, mixed-state receipt preservation/precedence, duplicate requirements, deterministic ordering, JSON round trip, and a 512 KiB request bound.

The four #194 semantic blobs were rebuilt on the current repository after #187 landed. Main advanced once during the first successful run, so that first pass was deliberately discarded as the final promotion authority. The same four blobs were then reanchored again onto:

```text
main: 0c7da4f3ad148c4d14bce127c75cfe4b03b658c8
head: 6706755e434a9bb533d77655587b01cbdd3fa1e8
commits: 1
files: 4
```

Exact-current ordinary CI:

```text
run: 32269578803
job: 96122443251
result: success
```

Generated-provenance review:

```text
run: 32269578744
job: 96122443145
result: success
```

Formatter, strict Clippy, modern project/review/closure/external-reference guards, active-work preflight, full tests, and repository/history/CI-filter/diff dogfood all passed. Main remained at the exact reanchor base through the gate.

The earlier executed receipt remains in `research/observation-frontiers.md`; this current-main pass proves the same frontier bytes compose with the present repository.

## Boundary / next

Phase A performs zero source acquisition, probe lookup, evidence-strength inference, or mapping from discriminator ID to #145 `{kind,target}`.

The next commit is intentionally the #201 negative control proving that a KNOWN value can still be labeled `Current` while shared applicability says INVALID or UNKNOWN. #210 then repairs that type contract.

Files: `src/observation_frontier.rs`, `tests/observation_frontier.rs`, `examples/observation_frontiers.rs`, `research/observation-frontiers.md`.

Refs #145 #148 #179 #185 #187 #190 #201 #210.

### DIFF
--- a/examples/observation_frontiers.rs
+++ b/examples/observation_frontiers.rs
@@ -1,0 +1,39 @@
+#![allow(dead_code)]
+
+use std::error::Error;
+use std::io::{self, Read};
+
+#[path = "../src/discriminator_observation.rs"]
+mod discriminator_observation;
+#[path = "../src/observation_frontier.rs"]
+mod observation_frontier;
+
+use observation_frontier::{
+    MAX_OBSERVATION_FRONTIER_REQUEST_BYTES, evaluate_observation_frontiers,
+    parse_observation_frontier_request,
+};
+
+fn main() {
+    if let Err(error) = run() {
+        eprintln!("observation-frontiers: {error}");
+        std::process::exit(1);
+    }
+}
+
+fn run() -> Result<(), Box<dyn Error>> {
+    let mut bytes = Vec::new();
+    io::stdin()
+        .take((MAX_OBSERVATION_FRONTIER_REQUEST_BYTES + 1) as u64)
+        .read_to_end(&mut bytes)?;
+    if bytes.len() > MAX_OBSERVATION_FRONTIER_REQUEST_BYTES {
+        return Err(format!(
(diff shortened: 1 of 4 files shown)


teamleaderleo:
Adversarial contract check before Phase B: `evaluate_requirement()` currently maps every same-subject `DiscriminatorValueState::Known` directly to frontier `Current`, while `applicability_ref` is only copied through and never interpreted. That is coherent only if #185's producer contract says KNOWN already means **known at the admitted current coordinate**. If value knowledge and applicability are separate axes, this can suppress a stale/unknown observation frontier as current. I opened #195 to force the split explicitly. Please add a moved-revision control proving either (A) the source adapter must convert stale known values to INVALID/UNKNOWN before this layer, or (B) the frontier consumes typed applicability state. An opaque applicability ref should not by itself make KNOWN current.

Issue #196, opened 2026-08-19
Pressure-test scoped JEI budget priority against decision-changing scope evidence


teamleaderleo:
## Trigger

#181/#189 established a real Stensibly discriminator:

```text
file-local convex/schema.ts packet
  -> misses #1571/#1573 even at 1 MiB

explicit convex/ scope history
  -> recovers both inside the 32 KiB envelope target
```

Draft #191 promotes explicit scope into Rust and currently uses this semantic eviction priority:

```text
scope_recent_history
  -> evict first

target-local packet detail
  -> evict afterward
```

That policy is locally conservative, but the Stensibly case is exactly one where the decision-changing evidence exists only in scope history.

## Research question

> Under budget pressure, can a blanket target-local-first policy delete the evidence that justified widening scope while preserving target-local history that did not answer the task?

## First adversarial fixture

Use the #191 scoped compiler semantics with:

```text
target-local recent history
  T1 harmless/file-local context

scope history
  S1 distributed failure lesson A
  S2 distributed failure lesson B
```

Define a tiny fixture-local next-action oracle:

```text
S1 + S2 visible -> inspect/reconcile distributed failure family
otherwise       -> continue file-local investigation
```

Choose a budget just below the candidate envelope so #191's current policy evicts one scope summary while leaving target-local recent history intact.

Require the research control to demonstrate:

```text
before budget -> distributed action
after budget  -> file-local action
```

while the packet truthfully receipts the scope eviction.

A passing negative control proves only that locality priority is not decision-preservation policy.

## External sweep

After the synthetic discriminator, replay the exact #189 Stensibly case at smaller budgets below the ~19.1 KiB candidate envelope (for example 16 KiB / 12 KiB / 8 KiB where feasible).

Record:

- first budget at which #1571/#1573 disappear;
- scope vs target-local eviction receipts;
- whether the scoped envelope still fits;
- whether protected core eventually fails closed.

Do not weaken the 32 KiB acceptance result.

### Current priority survives

If the real Stensibly lessons survive all meaningful budgets until the envelope fails closed, target-local-first remains a useful heuristic for this corpus.

### Current priority fails

Then the next selector needs a stronger rule than locality, such as:

- explicit protected scope items supplied by the caller;
- task-conditioned relevance/discriminator requirements;
- justification/frontier edges identifying which evidence can change the next action;
- whole-relation eviction instead of arbitrary chronological scope trimming.

Do not jump directly to a scalar relevance score.

## Boundary

[...]

PR #197, opened 2026-08-19
research: prove scoped budget locality can change next action

teamleaderleo:
Historical negative control; resolved by merged #282.

#197 earned the key counterexample:

```text
scope_recent_history-first byte eviction
can erase decision-changing distributed evidence
while preserving less-useful target-local history
```

#205 then proved exact upstream-selected scope refs can be preserved, and #282 promoted that seam into the existing scoped compiler with a durable pinned Stensibly replay.

Landed discriminator from #282:

```text
unprotected 15750 -> both selected lessons lost
protected   15750 -> both selected lessons preserved
protected    4096 -> both selected lessons preserved in 3911 bytes
protected    3900 -> fail closed at required floor 3911
```

Merged implementation: `ef94b1ec9e9b20335599ffcdec47b377f65acf50`.

This branch remains the executable historical proof that locality priority alone was insufficient; it no longer represents active work.

Refs #196 #205 #281 #282.

### DIFF
--- a/.github/workflows/scoped-budget-stensibly-pressure.yml
+++ b/.github/workflows/scoped-budget-stensibly-pressure.yml
@@ -1,0 +1,232 @@
+name: Stensibly scoped JEI budget pressure
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    paths:
+      - ".github/workflows/scoped-budget-stensibly-pressure.yml"
+      - "tests/scoped_budget_action_loss.rs"
+      - "examples/agent_context_packet.rs"
+      - "examples/scoped_agent_context_packet.rs"
+      - "examples/support/scoped_agent_context_packet_impl.rs"
+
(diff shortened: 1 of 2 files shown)


teamleaderleo:
Executed result: exact head `e8ffa78dd4bd3893f07a351e41846ee9766b3e8d` passed ordinary CI run `32249877846` (#1321), including format, strict Clippy, active-work preflight, full tests, and normal dogfood. The test imports #191's actual scoped compiler. Under a 200-byte pressure step it preserves target-local recent history, evicts one `scope_recent_history_summary`, and the fixture-local next action changes from `InspectDistributedFailureFamily` to `ContinueFileLocal`. This earns the narrow negative result: **locality priority is not a decision-preservation policy**. Next discriminator is the pinned Stensibly sweep below the real Rust candidate size (18,794 bytes).

PR #205, opened 2026-08-19
research: protect selected scope evidence from budget eviction

teamleaderleo:
Historical comparator superseded by merged promotion #282.

#205 earned the exact-ref protection seam on the Stensibly budget cliff. #282 promoted that seam into the existing scoped compiler, kept ordinary calls on the original #191 path, added an explicit selected-SHA receipt, and passed both merge-view CI and the pinned public replay.

Landed behavior:

```text
unprotected 15750 -> both selected lessons lost
protected   15750 -> both selected lessons preserved, 15721 bytes
protected    4096 -> both selected lessons preserved, 3911 bytes
protected    3900 -> fail closed at required floor 3911
```

Executed promotion receipt:

```text
run       32280602283
result    success
merge     ef94b1ec9e9b20335599ffcdec47b377f65acf50
```

This older PR remains useful historical evidence for the research comparator and the earlier 3,953-byte floor. The promoted wrapper measured a smaller 3,911-byte floor while retaining the exact SHA audit receipt.

Refs #196 #197 #281 #282.

### DIFF
--- a/.github/workflows/scoped-selected-evidence-floor.yml
+++ b/.github/workflows/scoped-selected-evidence-floor.yml
@@ -1,0 +1,119 @@
+name: Stensibly selected evidence floor
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    paths:
+      - ".github/workflows/scoped-selected-evidence-floor.yml"
+      - "examples/scoped_selected_evidence_protection.rs"
+      - "examples/agent_context_packet.rs"
+      - "examples/support/scoped_agent_context_packet_impl.rs"
+
+permissions:
+  contents: read
+
+jobs:
+  floor:
+    if: github.head_ref == 'research/scoped-selected-evidence-protection'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          path: cultist
+          fetch-depth: 1
+
+      - name: Install stable Rust
+        run: |
+          rustup toolchain install stable --profile minimal
+          rustup default stable
+
+      - name: Build protected scoped example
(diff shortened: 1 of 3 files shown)


teamleaderleo:
Executed external A/B and floor result.

Comparator run `32250810021`, artifact `9364332943` (`sha256:56439ffe58f709166af25ed067ad8454dbe8c301c6b4f043ded55acaa65e96f9`) reproduced the landed policy cliff and showed the exact-ref protected policy preserves #1571/#1573 at 15,750 bytes and every tested budget through 4,096.

Selected policy tradeoff at 4,096:

```text
selected bytes                  3,953
scope history rows                  2  (#1573/#1571)
scope evictions                  16/18
target-local recent history         0
file support examples evicted    16
file counterexamples evicted     16
file companion relations evicted  8
file recent history evicted      17
file companion exclusions evicted 1
```

Floor run `32251052173`, artifact `9364417419` (`sha256:50ba8b1128a74394eaff6cbeefc678496c168b6adb6d46e56d481ef88343b2c8`) then pushed below 4 KiB. Exact boundary:

```text
budget 4096 -> success, selected bytes 3953, both refs preserved
budget 3900 -> ProtectedCoreTooLarge(required=3953)
all lower tested budgets -> same required protected core 3953
```

This earns the narrow design rule: **once an upstream JEI selector marks exact evidence as required, byte budgeting may spend unselected detail but must not silently delete selected evidence. If selected evidence + protected core cannot fit, return budget-insufficient/fail closed.**

Important follow-up: the current research comparator can preserve selected scope refs by evicting `companion_exclusion_detail` because the landed file-packet policy still considers that class evictable. Selected-evidence protection must compose with independently earned protected-core classes; it is not a new universal priority that may erase other protected evidence.

No subject/recency/path-locality inference was needed in the budget layer.

Issue #220, opened 2026-08-19
Close the edit-class observation acquisition loop


teamleaderleo:
## Trigger

#216 now proves the generic Phase B bridge:

```text
noncurrent observation D@S
-> explicit source-owned ObservationProbeBridge
-> existing #145 planner
-> selected / blocked / unresolved probe plan
```

The next missing evidence is producer-specific. The retained #179/#185 Oxc family already has a real deterministic source discriminator from #54:

```text
edit_class
  syntax_changed
  comments_or_docs_only
  unclassified
```

and retained subject:

```text
oxc-project/oxc@8783524015b1e6ff1c39ccf426df0bb07cbbc588:
crates/oxc_linter/src/rules.rs
```

## Experiment

Build one research adapter that reuses the existing `rust_syntax_cohort` first-parent token classifier and closes the loop:

```text
MISSING edit_class @ exact Oxc subject
-> source-owned bridge
-> existing #216 + #145 planner selects exact read-only edit-class probe
-> execute existing classifier at exact checkout head/path
-> emit v2 DiscriminatorObservation
-> v2 frontier becomes CURRENT
```

The source adapter should own:

- exact repository/revision/path subject identity;
- the mapping from `edit_class` to its probe `DiscriminatorKey`;
- exact clearing requirements;
- current-head verification;
- conversion of the existing classifier result into the v2 value reference;
- exact source/applicability receipts.

The generic bridge/planner/frontier should remain unchanged.

## Controls

- syntax-changing Rust edit -> KNOWN `syntax_changed` + APPLIES;
- comment/doc/whitespace-only edit -> KNOWN `comments_or_docs_only` + APPLIES;
- unclassifiable revision/file -> UNKNOWN current value, never a guessed partition;
- checkout head mismatch fails or yields explicit noncurrent evidence; it cannot claim APPLIES;
- similarly named probe without the source bridge remains unmapped;
- wrong subject/path cannot satisfy the bridge;
- executing the selected deterministic probe and inserting its produced observation changes the exact frontier from MISSING/UNKNOWN to CURRENT;
- retained Oxc `8783524…:rules.rs` carrier reproduces `syntax_changed` from the real public repository;
- source execution grants no broader evidence strength or action authority.

## Reuse boundary

The existing #54 classifier is the owning discriminator. Reuse it directly; do not introduce a second Rust syntax classifier in the adapter.

## Success criteria

[...]

PR #210, opened 2026-08-19
research: separate discriminator knowledge from applicability

teamleaderleo:
Closes #195 by replacing the v1 frontier contract demonstrated by #201 with separate typed value-knowledge and current-applicability axes.

## Before / after

#201 is now on `main` and proves v1 can represent:

```text
KNOWN old value
+ applicability INVALID | UNKNOWN
-> frontier CURRENT
```

This carrier chooses Model B:

```text
value_state
  known { value_ref }
  unknown { reason_ref }

applicability
  status = applies | unknown | invalid
  receipt_ref
```

A shared generic classifier combines the supplied axes:

```text
INVALID applicability -> invalid, preserve known old value
UNKNOWN applicability -> unknown, preserve known old value
APPLIES + value UNKNOWN -> unknown
APPLIES + value KNOWN -> current
```

Generic code still executes no source-domain applicability logic. Source adapters supply typed applicability plus exact receipt.

Because admitted wire meaning changes, discriminator-observation and observation-frontier research schemas advance from v1 to v2. The retained three-family observation corpus carries explicit `applicability.status=applies` receipts.

## Closure controls

- every selected #179 discriminator remains current only through KNOWN+APPLIES;
- KNOWN + UNKNOWN applicability -> UNKNOWN frontier, known value retained;
- KNOWN + INVALID applicability -> INVALID frontier, known value retained;
- value UNKNOWN + APPLIES -> UNKNOWN;
- missing applicability receipt rejects;
- duplicate/conflict, source identity, wrong-subject, mixed-state, ordering, round-trip, and bounds controls remain;
- the exact #201 moved-head and missing-current-revision counterexamples are inverted into closure tests using the real shared applicability evaluator.

## Current-main promotion receipt

After #201 landed as `592111d1d747b4f89e900ea788adfd1a9607b445`, the seven repair blobs were flattened to one commit directly on that exact main tree:

```text
head: db78d92649a1df07ee0e7bf9b7410b7102ab157c
commits: 1
files: 7
```

Ordinary CI:

```text
run: 32270894042
job: 96126771850
result: success
```

Generated-provenance review:

```text
run: 32270894038
job: 96126771516
result: success
```

Formatter, strict Clippy, current repository policy guards, active-work preflight, all v2/closure tests, and repository/history/CI-filter/diff dogfood passed. Main remained at the exact base through the gate.

The earlier semantic receipt remains in `research/typed-observation-applicability.md`; this is the current-main closure receipt.

## Known source-data boundary

[...]

Issue #145, opened 2026-08-19
Research evidence acquisition planning from material UNKNOWNs


teamleaderleo:
## Thesis

JEI asks which evidence a task needs. Performance work asks how to make evidence collection cheap. Durable UNKNOWN/frontier work asks what is still missing.

The missing bridge is an execution question:

> Given a material UNKNOWN, which deterministic evidence probe is capable of resolving it, and what is the least expensive justified probe to run next?

Cultist analyzers already have very different evidence costs: exact local facts, source parses, Git history, semantic adapters, provider snapshots, target execution, and optional remote enrichment. A research planner can make those costs and capabilities explicit without turning Cultist into an opaque optimizer.

## Candidate probe contract

```text
Probe
  id
  prerequisites
  can_establish / can_refute
  evidence class produced
  applicability dimensions required
  cost vector
  effect boundary
```

Example:

```text
probe generated_header
  cost: files_opened=1
  can_establish: generated(output)

probe generator_declaration
  cost: files_parsed=1
  can_establish: owns(generator,input,output)

probe historical_companion
  cost: git_history_query=1
  can_establish: OBSERVED co-change only

probe semantic_resolution
  cost: semantic_adapter_calls=1
  can_establish: resolved call/reference relation
```

## Cost representation

Avoid one magic score in v0. Preserve an inspectable vector such as:

```text
git_processes
files_opened
files_parsed
history_commits
remote_requests
bytes
semantic_adapter_calls
effectful_executions
```

Selection can begin with an explicit lexicographic/product policy supplied by the research fixture.

## First experiment

Take three UNKNOWN species:

1. generated ownership unknown;
2. exact-head validation missing;
3. current provider/work coordinate missing.

For each:

- enumerate admitted probes;
- prove some probes are incapable of satisfying the obligation despite being cheap;
- select the cheapest capable probe under an explicit policy;
- execute or simulate the deterministic result;
- recompute the obligation;
- stop once the required discriminator is established or all admitted probes are exhausted.

## Boundaries

- Planning never strengthens the evidence a probe can produce. Historical co-change stays OBSERVED even if it is cheap.
- Effectful probes remain explicitly separated and require caller/orchestrator authority.
- Provider/model/network availability is evidence/cost state, not silently assumed.
- A planner may return `no admitted probe can resolve this obligation`.
- Probe capabilities should come from producer/analyzer definitions, not prose keyword inference.

## Composition

- #49 performance / demand-driven work;
[...]

Issue #142, opened 2026-08-19
Research counterexample-driven cohort refinement


teamleaderleo:
## Thesis

Cultist already preserves counterexamples so repository frequency does not silently become policy. The next useful step is to ask whether a counterexample reveals the missing discriminator of an otherwise useful cohort.

Research question:

> Given a precedent/association with support and counterexamples, can Cultist deterministically discover a smaller observable cohort in which the relation becomes more precise?

Example:

```text
all comparable A changes
  14 / 15 update B

counterexample
  one docs-only A change omits B

candidate refinement
  semantic A changes: 14 / 14 update B
  docs-only A changes: 0 / 1 update B
```

The goal is not to synthesize universal rules. Counterexamples should teach candidate scope/cohort boundaries that remain inspectable and replayable.

## Candidate discriminator vocabulary

Keep v0 intentionally bounded to repository-visible facts already available or cheap to derive:

- exact/prefix path scope;
- package/crate/module;
- test vs production;
- generated vs authored;
- docs-only vs code-bearing change;
- cfg/target family where explicit;
- recent vs historical cohort;
- explicit guidance/decision present;
- change breadth bucket;
- analyzer-specific deterministic fact already present in the evidence packet.

No arbitrary prose classification in v0.

## First deterministic experiment

Use one historical companion fixture with a known meaningful counterexample and one precedent fixture with genuine mixed scope.

Algorithm sketch:

1. start from the full support/opportunity cohort;
2. enumerate one discriminator at a time from the admitted vocabulary;
3. partition support + counterexamples;
4. retain candidate refinements only when they explain at least one counterexample and preserve useful positive support;
5. emit exact counts and the discriminator definition;
6. keep ties/multiple explanations explicit;
7. reject refinements that merely memorize one path/commit without reusable meaning.

## Adversarial controls

- random/noisy counterexample with no admitted discriminator -> remain mixed/UNKNOWN;
- discriminator that improves purity by isolating one single commit -> reject as overfit in v0;
- two competing refinements -> surface both;
- current change lacks the discriminator fact -> applicability of the refinement remains UNKNOWN;
- temporal adjacency alone never becomes a causal cohort.

## Output concept

```text
OBSERVED
  14 / 15 comparable changes update B

COUNTEREXAMPLE
  commit X omitted B

CANDIDATE COHORT
  docs-only changes: 0 / 1 update B
  code-bearing changes: 14 / 14 update B

CURRENT CHANGE
  code-bearing

[...]

PR #221, opened 2026-08-19
research: close the Rust edit-class observation loop

teamleaderleo:
Progresses #220 as the first real source-adapter child of green #216.

## Closed loop

This carrier reuses #54's existing Rust first-parent token classifier and exercises:

```text
MISSING edit_class @ exact focused repo/revision/path
-> source-owned bridge
-> existing #216 / #145 planner selects read-only rust_edit_class probe
-> existing classifier executes
-> v2 source observation
-> exact frontier becomes CURRENT
```

The generic bridge, planner, applicability evaluator, and frontier modules remain unchanged.

## Focused-commit correction

The first public run exposed a hidden #54 precondition. Pinned repository head `8783524015b1e6ff1c39ccf426df0bb07cbbc588` does not change `crates/oxc_linter/src/rules.rs`. Raw token equality there initially looked like `comments_or_docs_only`, while #54 only classified non-merge commits selected by `git log -- rules.rs`.

The adapter now requires exactly one parent plus an actual change to the exact anchor path before emitting a known edit class. Repository head `8783524…` is retained as a negative control and emits UNKNOWN `anchor-unchanged` + APPLIES.

The latest actual focused `rules.rs` commit under that pinned history is:

```text
```

That exact commit reproduces KNOWN `syntax_changed` + APPLIES.

## Source contract

`RustEditClassSubject` binds repository, exact 40-hex revision, and normalized repository-relative Rust path. Collection emits one source-owned bridge, one read-only probe definition, and one v2 `DiscriminatorObservation`.

Focused classifier states map only as syntax_changed / comments_or_docs_only / UNKNOWN. Root/merge/anchor-unchanged revisions are UNKNOWN before classification. Shared applicability binds repository + exact revision + exact path to the checkout's actual HEAD; a moved checkout preserves an old known value only as INVALID current evidence.

## Public and local controls

Local Git fixtures cover syntax change, comment-only change, root UNKNOWN, unrelated-anchor UNKNOWN, moved-head INVALID, v2 round-trip, admission failures, and full MISSING -> selected probe -> produced observation -> CURRENT composition.

The public Oxc workflow preserves both earned coordinates:

```text
pinned head 8783524…
  -> UNKNOWN anchor-unchanged

focused commit 228e8e0f85c0…
  -> KNOWN syntax_changed
  -> applicability APPLIES
  -> read-only probe SELECTED
  -> final frontier CURRENT
```

The inherited #185 observation corpus is corrected on this branch to the exact focused subject and workflow receipt.

## Validation

The exploratory public run `32257281636` / #1 exposed the arbitrary-head precondition bug. After the focused-commit repair, public run `32257998359` / #7 succeeded; the corrected corpus then passed ordinary CI #1533 and public carrier #9.

The branch was compacted to one semantic child commit:

```text
08d4521a22baa250c38d87852af7bfc611ac39bb
```

on #216 head `2f20ef16fb05313b1df4cf73cee913dce19b8318`.

That exact compacted head passed full CI run `32258598148` / #1541 and public Oxc carrier run `32258599172` / #11.

## Reuse boundary

The only #54 producer edit is visibility for `RustEditClass` and `classify_rust_edit`; token/fingerprint/classification semantics are unchanged. Generic #216 modules are unchanged.

Refs #54 #142 #145 #168 #179 #185 #190 #210 #216 #220.

### DIFF
--- a/examples/rust_edit_class_observation.rs
+++ b/examples/rust_edit_class_observation.rs
@@ -1,0 +1,68 @@
+use std::env;
+use std::error::Error;
+use std::path::{Path, PathBuf};
+
+#[allow(dead_code)]
+#[path = "../src/applicability.rs"]
+mod applicability;
+#[allow(dead_code)]
+#[path = "../src/discriminator_observation.rs"]
+mod discriminator_observation;
+#[allow(dead_code)]
+#[path = "../src/durable_obligation.rs"]
+mod durable_obligation;
+#[allow(dead_code)]
+#[path = "../src/evidence_planner.rs"]
+mod evidence_planner;
+#[allow(dead_code)]
+#[path = "../src/justification.rs"]
+mod justification;
+#[allow(dead_code)]
+#[path = "../src/observation_frontier.rs"]
+mod observation_frontier;
+#[allow(dead_code)]
+#[path = "../src/observation_probe_bridge.rs"]
+mod observation_probe_bridge;
+#[allow(dead_code)]
+#[path = "../src/rust_edit_class_source.rs"]
+mod rust_edit_class_source;
+
+use rust_edit_class_source::{RustEditClassSubject, collect_rust_edit_class_source};
+
+fn main() {
+    if let Err(error) = run() {
+        eprintln!("rust-edit-class-observation: {error}");
+        std::process::exit(1);
+    }
+}
+
+fn run() -> Result<(), Box<dyn Error>> {
+    let mut args = env::args().skip(1);
+    let root = PathBuf::from(
+        args.next()
+            .ok_or("usage: rust_edit_class_observation REPO_ROOT REPOSITORY REVISION FILE")?,
+    );
+    let repository = args
+        .next()
+        .ok_or("usage: rust_edit_class_observation REPO_ROOT REPOSITORY REVISION FILE")?;
+    let revision = args
+        .next()
+        .ok_or("usage: rust_edit_class_observation REPO_ROOT REPOSITORY REVISION FILE")?;
+    let path = args
+        .next()
+        .ok_or("usage: rust_edit_class_observation REPO_ROOT REPOSITORY REVISION FILE")?;
+    if args.next().is_some() {
+        return Err("usage: rust_edit_class_observation REPO_ROOT REPOSITORY REVISION FILE".into());
+    }
+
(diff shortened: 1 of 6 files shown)


teamleaderleo (commented):
Rook 🦉 review of exact head `5a15c13fb361de7a101591f097f1c4055cace346`.

The focused edit-class/source behavior is otherwise coherent: single-parent + anchor-changed admits classification; root/merge-like and anchor-unchanged cases become value UNKNOWN; a known old classification remains visible with applicability INVALID after checkout HEAD moves; exact subject/path admission is bounded; the probe is read-only; and the produced observation closes only its exact frontier. Exact-head CI `32274991117` and generated provenance `32274991150` are green.

One blocker remains in the **generic source adapter applicability contract**: repository applicability is self-supplied rather than independently established from `root`.

`collect_rust_edit_class_source` builds:

```text
requirements.repository = subject.repository
context.repository      = subject.repository
```

while only `context.revision` is actually read from the checkout (`git rev-parse HEAD`). The function never proves that the canonicalized `root` is the repository named by `subject.repository`. A caller can therefore point the adapter at another Git repository that contains the requested revision/path and still receive repository APPLIES because the same caller-provided repository string is copied to both sides of #123. The current tests use `owner/repo` for temporary repositories and contain no wrong-repository control, so they reproduce the self-fulfilling match.

The retained hosted Oxc receipt is still credible at its **workflow** boundary because that workflow explicitly checks out `oxc-project/oxc` before invoking the adapter. The generic `collect_rust_edit_class_source` API itself has not earned repository APPLIES without that outer binding.

Smallest repair options:

- accept an independently supplied/verified current repository identity alongside `root` and feed that into `EvaluationContext.repository`; or
- make repository identity an explicit outer precondition/receipt and omit repository from this function's applicability dimensions rather than claiming it was observed locally.

Add a wrong-repository control before promoting this source adapter as a reusable currentness authority. Historical #235/#287 Oxc proofs can continue citing the hosted workflow binding separately.

teamleaderleo:
Superseded on current main by merged #339, which preserves the Rust edit-class classifier/focus semantics while repairing the repository applicability self-binding with an independently supplied current checkout repository coordinate. The selected Oxc end-to-end composition then passed and merged through #335. Closing this historical draft carrier; its original public receipts remain useful history.

Issue #225, opened 2026-08-19
Mine Agent Skills evolution and failures as a capability-demand research corpus


teamleaderleo:
## Thesis

Public Agent Skills repositories are a uniquely valuable feeder corpus for Cultist's worker-capability / JEI research.

A skill is essentially a version-controlled attempt to externalize procedural cognition:

```text
repeated task burden
-> instructions / scripts / references
-> on-demand selection
-> agent execution
-> issue / failure / refinement
```

That is extremely close to #106/#215's longitudinal question:

> Which cognition can be moved out of repeated model reasoning and into durable, reusable computation or context?

Mine **skill evolution, bug reports, eval failures, and maintenance changes** to discover candidate failure species and discriminators. Do not treat Anthropic/OpenAI/other skill conventions as Cultist product policy.

Refs #17 #29 #41 #62 #106 #115 #137 #146 #148 #215 #223.

## Why this corpus is unusually useful

The Agent Skills format is now used publicly across multiple agent ecosystems. Skills package:

- activation descriptions;
- operating instructions;
- scripts;
- reference material;
- templates/resources;
- environment assumptions.

Their maintenance history therefore exposes several boundaries that ordinary code benchmarks conflate:

```text
skill not selected
skill selected but wrong content loaded
skill content too broad / expensive
instructions ambiguous
script path/environment wrong
reference stale
validator/eval harness wrong
agent misinterprets correct skill
skill works but only under one platform/toolchain
```

A final failed task does not tell these apart. Skill repositories often preserve enough source + issue evidence to do so.

## External scale signal

Recent research suggests this is not a tiny niche corpus:

- **GitSkills (Aug 2026)** reports 3,797,117 `SKILL.md` files from 282,200 public GitHub repositories, grouped into 1,877,981 distinct contents, with repository/path/hash metadata and some commit history.
  - paper: https://arxiv.org/abs/2608.10906
- **From Registry to Repository (Jul 2026)** studies 18,463 registry skills plus 23,199 personal-use skills from 5,876 GitHub repositories and reports that operational specifications and knowledge/resources are major targets of change.
  - paper: https://arxiv.org/abs/2607.00911
- Open Agent Skills specification / examples:
  - https://github.com/agentskills/agentskills
- Anthropic public skill implementation/issues:
  - https://github.com/anthropics/skills
- OpenAI Codex skill catalog/issues:
  - https://github.com/openai/skills

[...]

Issue #223, opened 2026-08-19
Research metamorphic capability probes from evidence mutations


teamleaderleo:
## Thesis

Model/worker "skills" are easier to study when the evaluation varies one repository-evidence distinction at a time instead of inferring capability from one end-to-end patch outcome.

Compose #146 evidence mutation testing with #215 worker capability-demand decomposition:

```text
canonical task + envelope E
        ↓
mutate exactly one admitted semantic dimension
        ↓
envelope E'
        ↓
same worker / harness / repository task
        ↓
compare next justified action
```

The test oracle is a **metamorphic relation**:

- some mutations should force the action to change;
- some should leave the action invariant;
- some should force `UNKNOWN` / escalation;
- no private chain-of-thought is needed.

The product question is:

> Which exact evidence distinctions can a worker reliably respond to once repository search, task identity, and tool affordances are held fixed?

Refs #106 #115 #123 #134 #137 #146 #173 #195 #201 #210 #215 #218.

## Why this is cleaner than ordinary skill labels

A final task failure conflates:

```text
repository exploration
JEI selection
attention
semantic interpretation
planning
execution
validation
harness affordances
specification quality
```

A metamorphic pair can isolate one candidate capability demand.

Example:

```text
same task
same report
same worker
same tools

A
  value = KNOWN(x)
  applicability = APPLIES

B
  value = KNOWN(x)
  applicability = INVALID

required relation
  action(B) must not treat x as current
```

Failure on B after A succeeds is evidence about **applicability sensitivity under this fixed condition**, not generic coding ability.

## First capability families

Use only semantics whose deterministic owner/oracle already exists.

### 1. Applicability / freshness sensitivity

Compose #123/#201/#210.

```text
KNOWN + APPLIES
  -> may use current value

KNOWN + INVALID
  -> must not use as current

KNOWN + UNKNOWN
  -> current use unsupported; request/recover missing context
```

Invariant control:

```text
move an unrequired coordinate
  -> action should remain unchanged
```

This gives both sensitivity and specificity controls.

### 2. Counterexample sensitivity

Compose #125/#146/#173.

```text
support-only detail removed
  -> action may remain invariant in an earned fixture

material counterexample removed
  -> canonical action relation must expose that the stronger action is no longer justified by the same evidence
```

When using a worker, compare the canonical and mutated envelopes against the deterministic action oracle. Do not assume "more cautious" is always correct; unnecessary escalation is also a miss.

[...]

Issue #215, opened 2026-08-19
Research worker capability demand by decomposing agent failures


teamleaderleo:
## Thesis

Coding-agent failure is not a single signal.

A failed task can arise because:

- required repository evidence did not exist;
- Cultist had the evidence but JEI failed to select it;
- the packet contained the evidence but the worker never inspected it;
- the worker inspected it but interpreted it incorrectly;
- the worker reached the right conclusion but chose the wrong plan;
- the plan was right but tool use / implementation failed;
- the implementation was plausible but validation/completion discipline failed;
- the task or environment withheld a necessary affordance;
- the worker correctly detected a capability/evidence boundary and escalated.

Those cases look similar in an end-to-end pass/fail benchmark and imply completely different product work.

Research whether Cultist can preserve enough typed episode evidence to distinguish **repository-legibility failures, harness/affordance failures, and residual worker capability demands**.

The goal is not a leaderboard or permanent model taxonomy.

The question is:

> After holding the task, repository state, evidence envelope, tools, and completion contract as fixed as practical, what capability did the worker still have to supply itself?

Refs #41 #62 #74 #99 #106 #115 #137 #146 #148.

## Failure decomposition

Start with a small episode vocabulary. These are candidate diagnostic outcomes, not labels to infer from prose alone.

```text
EVIDENCE_ABSENT
  the needed fact/discriminator was not available from the admitted repository/evidence sources

SELECTION_MISS
  the needed evidence existed in the candidate evidence set but was omitted from the worker packet

ATTENTION_MISS
  decisive evidence was present in the supplied packet but the trajectory never inspected/used it

INTERPRETATION_MISS
  decisive evidence was inspected but the worker drew a conclusion inconsistent with the typed evidence / deterministic oracle

PLANNING_MISS
  worker reached the relevant conclusion but selected an action that did not satisfy the task/stop/completion contract

EXECUTION_MISS
  plan was sufficient but implementation/tool use failed to realize it

VALIDATION_MISS
  candidate implementation existed but required validation was skipped, misread, or treated as stronger than it was

AFFORDANCE_MISS
  worker lacked a required tool/probe/environment capability even though the task demand was otherwise explicit

SPECIFICATION_GAP
  the task itself omitted a constraint that the evaluator later required and repository evidence did not establish

CORRECT_ESCALATION
  worker correctly identified that evidence/capability was insufficient and stopped or requested the exact missing discriminator

SUCCESS
  bounded task/completion contract satisfied
```

[...]

Issue #146, opened 2026-08-19
Research evidence mutation testing for JEI and terse projections


teamleaderleo:
## Thesis

Cultist is beginning to intentionally omit evidence from initial agent/review projections. The key safety question is behavioral:

> Which evidence can be removed, weakened, made stale, or misbound before the receiver chooses a different or unjustified next action?

Borrow the testing philosophy of mutation testing, but mutate evidence packets instead of source programs.

## Candidate mutations

Start with typed research fixtures, never prose keyword rewrites:

```text
drop supporting receipt
drop counterexample
drop applicability limit
move exact revision
swap work identity
remove omission count
remove clearing condition
replace APPLIES with missing context
reorder report-local positional refs
remove source/authority receipt
```

Each mutation should target one semantic distinction already represented by Cultist research.

## Oracle

Every fixture declares the justified next disposition/question from the full canonical object, for example:

```text
HOLD
NEXT run target test at exact HEAD
```

Then compare the mutated projection.

A mutation is decision-relevant when it causes one of:

- stronger conclusion than canonical evidence allows;
- weaker/unnecessary escalation;
- different next discriminator;
- hidden material UNKNOWN;
- stale evidence accepted as current;
- wrong object/reference updated;
- missing expansion request that canonical evidence requires.

## First corpus

Compose existing adversarial fixtures from:

- #119/#121 terse projection;
- #123/#124 applicability;
- #125 evidence roles;
- #127/#128 stable identity;
- #62/#67 JEI omission receipts;
- active-work freshness/coordination research.

Example:

```text
canonical
  target test passed
  LIMIT Linux only
  -> NEXT obtain required cross-platform evidence

mutation: drop LIMIT
  -> consumer treats PASS as sufficient
  => unsafe omission
```

## Measures

- initial bytes/tokens;
- mutation survived / killed by decision oracle;
- wrong-action mutations;
- unnecessary-expansion mutations;
- evidence role/field responsible;
- cross-model variance when an agent consumer is included;
- deterministic baseline behavior independent of models.

## Product use

This harness can help decide:

- whether #125 evidence roles deserve canonical fields;
- what terse/JEI output may omit;
- which omission receipts belong in the initial packet;
- which applicability/identity fields are material;
- whether a proposed compact encoding lost a semantic distinction.

## Boundary

- The oracle is a fixture/replay decision contract, not an automatic merge authority.
- Mutations operate on typed semantic objects where possible.
[...]

PR #226, opened 2026-08-19
research: seed agent skill failure corpus with explicit claim status

teamleaderleo:
Starts #225 with a small, typed **reported-only** external corpus and ordinary admission test.

## Why

Agent Skills issue trackers contain unusually direct examples of procedural knowledge being packaged, selected, executed, measured, and repaired. They are useful research feeders only if Cultist does not silently turn issue reports into facts.

## Seed

Adds `research/agent-skill-corpus-seed.json` with 14 public issue identities across:

```text
anthropics/skills
openai/skills
```

Candidate reported boundaries:

```text
measurement_misrepresented
selection_budget_overreach
context_binding_miss
affordance_miss
provenance_identity_gap
schema_tooling_drift
```

Every row is explicitly:

```text
evidence_status = reported_only
```

and carries a short **candidate discriminator** describing what would need to be tested before the episode can support stronger attribution.

## Examples in the seed

- trigger/eval harness reports where zero recall may reflect an unreachable/cross-matched skill instead of model selection failure;
- eager/duplicate skill payload reports as JEI/context-budget pressure;
- helper-path/root ambiguity as ambient-context pressure;
- Windows/dependency assumptions as affordance pressure;
- origin/install-root/spec drift as identity/applicability pressure.

Issue titles and URLs are source identities; issue bodies are not copied into the corpus.

## Admission test

`tests/agent_skill_corpus_seed.rs` requires:

- schema v1;
- 12..64 bounded rows;
- unique episode IDs;
- both sampled ecosystems present;
- at least five failure species;
- every current row remains `reported_only`;
- source URL namespace matches ecosystem;
- issue number in source URL matches the typed episode ID;
- non-empty title and candidate discriminator.

This test validates the research carrier, **not the vendor issue claims**.

## Next evidence states

Do not mutate `reported_only` by editorial confidence. A future episode should advance only with explicit evidence such as:

```text
reported_only
-> reproduced
-> repair_linked
-> counterfactual_replayed
```

Exact vocabulary remains research-local until real episodes need it.

## Boundary

- research corpus only; no product CLI/analyzer behavior;
- external repos are feeders, not specifications;
- no vendor-specific trigger semantics promoted to Cultist rules;
- no private skill data;
- no claim that issue author diagnosis is correct;
- provenance/security separate from authority;
- candidate boundary is a hypothesis until reproduced.

Refs #29 #41 #106 #115 #123 #134 #137 #146 #148 #215 #223 #225.

### DIFF
--- a/research/agent-skill-corpus-seed.json
+++ b/research/agent-skill-corpus-seed.json
@@ -1,6 +1,6 @@
 {
   "schema_version": 1,
-  "description": "Reported public Agent Skills issue cases for capability-demand research. Rows are claims until independently reproduced or linked to a verified repair.",
+  "description": "Public Agent Skills issue cases for capability-demand research. Evidence status records how far each reported diagnosis has been independently supported; it does not promote vendor issue claims by editorial confidence.",
   "episodes": [
     {
       "id": "anthropics/skills#556",
@@ -17,8 +17,12 @@
       "source_url": "https://github.com/anthropics/skills/issues/1352",
       "title": "skill-creator: run_eval.py parallel workers cross-match skill UUIDs, producing false-negative trigger rates",
       "reported_boundary": "measurement_misrepresented",
-      "evidence_status": "reported_only",
-      "candidate_discriminator": "compare parallel and isolated/serial trigger evaluation under the same skill description"
+      "evidence_status": "source_mechanism_confirmed",
+      "source_evidence_url": "https://github.com/anthropics/skills/blob/0a64e398ec6bb34a494f0c347e8ccae53a862f8e/skills/skill-creator/scripts/run_eval.py",
+      "source_evidence_blob_sha": "e58c70bea39d5b252a1e819f242bbdcdf20e8b87",
+      "repair_url": "https://github.com/anthropics/skills/pull/1437",
+      "repair_status": "proposed_open",
+      "candidate_discriminator": "compare parallel and isolated/serial trigger evaluation under the same skill description; current source confirms shared project registration, per-query UUID names, parallel execution, and exact per-query UUID matching, but the claimed trigger-rate consequence is not independently replayed here"
     },
     {
       "id": "anthropics/skills#1487",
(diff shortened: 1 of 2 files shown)


teamleaderleo:
Exact-head seed validation + first evidence graduation:

```text
head  ec3c107ad4d2e14371bbbb9a5a36d412474d16c6
CI    32258885962 / #1554 success
provenance dogfood 32258885925 / #291 success
```

Current corpus:

```text
14 public issue episodes
2 ecosystems
6 candidate reported-boundary species
```

Evidence status remains explicit per row.

### Graduated: `anthropics/skills#1352`

```text
reported_only
-> source_mechanism_confirmed
```

Exact source coordinate:

```text
repo   anthropics/skills
commit 0a64e398ec6bb34a494f0c347e8ccae53a862f8e
path   skills/skill-creator/scripts/run_eval.py
blob   e58c70bea39d5b252a1e819f242bbdcdf20e8b87
```

The source visibly contains the ingredients reported by the issue: per-query UUID command names, one shared project `.claude/commands` directory, parallel `ProcessPoolExecutor` queries against that same project root, and exact matching of each worker's generated UUID name. Open repair PR `anthropics/skills#1437` is linked as `proposed_open`.

This **does not** claim the reported trigger-rate consequence was independently replayed.

### Graduated: `openai/skills#96`

Exact source coordinates:

```text
repo   openai/skills
commit 49f948faa9258a0c61caceaf225e179651397431

SKILL.md
blob 0ee19e080dc9f0b082157080d1e65511ddc49e7e

scripts/fetch_comments.py
blob 09b9c01c50d9202bd6455ce36887e08be92ac1ae
```

The current skill tells the agent to `Run scripts/fetch_comments.py` without binding that relative path to the skill root, while the helper is bundled under the skill directory. This confirms an unbound relative-reference hazard in the source package.

It **does not** prove the reported Codex runtime failure or a universal cwd rule. No matching repair PR was found in the sampled search, so no repair lineage is invented.

### Admission discipline

The ordinary test now requires:

- exact commit-qualified source URLs for every source-confirmed row;
- 40-hex source blob identities;
- any repair record to use the matching vendor PR namespace and explicit status;
- `reported_only` rows to carry no stronger source/repair evidence;
- at least one source-confirmed episode in **each** sampled ecosystem.

Next evidence state should be earned by replay/repair evidence, not editorial confidence. The seed carrier is now good enough to freeze while #225 expands/refines the corpus elsewhere.

Refs #115 #134 #137 #146 #148 #215 #223 #225.

Issue #227, opened 2026-08-19
Bind refinement discriminator requirements to exact subjects


teamleaderleo:
## Trigger

#221 closes one real acquisition loop and makes the subject boundary concrete:

```text
edit_class @ oxc-project/oxc@228e8e0f...:rules.rs
```

is a current source observation, while pinned repository head `8783524…:rules.rs` is a different subject and correctly stays UNKNOWN because the anchor did not change.

The refinement ledger in #179 still expresses candidate needs as:

```text
CandidateRefinement.discriminator_refs: Vec<String>
```

For the Oxc selected transition this is only:

```text
["edit_class"]
```

There is no exact observation subject in the candidate requirement.

Meanwhile #194/#210/#216/#221 earned exact currentness/acquisition on:

```text
(discriminator_id, subject_ref)
```

## Counterexample

Use the retained Oxc selected refinement and remove its exact current observation.

Insert a different current observation:

```text
edit_class = syntax_changed
subject = unrelated/repo-or-commit/path
applicability = APPLIES
```

Then compare the two existing compositions:

### ID-only refinement coverage

The current #187 cross-object test indexes availability by `discriminator_id` only. It can therefore report that the selected Oxc candidate has a current `edit_class` observation.

### Exact frontier

Ask #210/#194 for:

```text
edit_class @ oxc-project/oxc@228e8e0f...:rules.rs
```

The same batch must produce:

```text
MISSING
other_subject = [unrelated edit_class observation]
```

If both are true, the current schemas permit a wrong-subject observation to satisfy refinement coverage even though the exact observation frontier says the required source coordinate is missing.

## Research question

Where should exact subject binding live?

Candidate models:

1. **Ledger-owned requirement**

```text
CandidateRefinement.discriminator_requirements[]
  discriminator_id
  subject_ref
```

2. **Separate source-owned requirement mapping**

Keep the generic refinement ledger source-agnostic and require an explicit receipt that maps one candidate discriminator ref to one exact observation subject before currentness/acquisition is evaluated.

The experiment should choose based on the retained families, not convenience.

## Controls

- exact Oxc observation satisfies the Oxc requirement;
- same `edit_class` on another subject cannot satisfy it;
- project-memory refinement can retain its episode-local subject identity without inventing repository coordinates;
- justification refinement can retain its obligation-local subject identity;
- multiple subjects for one discriminator remain inspectable;
- no implicit “first current observation for D” selection;
[...]

PR #228, opened 2026-08-19
research: prove refinement discriminator subject ambiguity

teamleaderleo:
Starts #227 as a one-file negative control on the green #221 source-adapter stack.

## Counterexample

The retained Oxc selected refinement in #179 requires only:

```text
discriminator_refs = ["edit_class"]
```

The test removes its earned current observation at:

```text
edit_class @ oxc-project/oxc@228e8e0f85c0e7aeded02c5e27fd810004d3b41a:crates/oxc_linter/src/rules.rs
```

and inserts a current KNOWN+APPLIES `edit_class=syntax_changed` observation for another subject (`other.rs`).

Then it composes the two existing semantics:

1. current #187-style cross-object coverage, collapsed by `discriminator_id`, still reports the selected Oxc candidate's `edit_class` as available;
2. exact #210 frontier for the earned `rules.rs` subject reports `MISSING` and preserves the wrong observation under `other_subject`.

A second control keeps the real corrected #221 subject CURRENT.

## Result

This ambiguity is executable. Formatted head `b046c0aadc617a0a341ae12c943c7ec7d28b5f11` passed full CI run `32259025041` / #1560: format, strict Clippy, active-work preflight, both subject-identity tests, and repository/history/CI-filter/diff dogfood all succeeded.

The same observation batch therefore admits both statements simultaneously:

```text
ID-only selected-refinement coverage: edit_class available
exact Oxc D@S frontier: MISSING
wrong edit_class observation: other_subject
```

## Boundary / repair direction

No repair lives in this carrier. #227 will use a separate source-owned candidate→observation requirement receipt rather than adding one intrinsic subject to `edit_class`: the Oxc forward/reverse experiments reuse the same discriminator across many commit subjects, so subject identity belongs to one candidate/application requirement, not to the reusable discriminator definition itself.

Refs #148 #179 #185 #187 #190 #194 #210 #216 #220 #221 #227.

### DIFF
--- a/examples/rust_edit_class_observation.rs
+++ b/examples/rust_edit_class_observation.rs
@@ -1,0 +1,68 @@
+use std::env;
+use std::error::Error;
+use std::path::{Path, PathBuf};
+
+#[allow(dead_code)]
+#[path = "../src/applicability.rs"]
+mod applicability;
+#[allow(dead_code)]
+#[path = "../src/discriminator_observation.rs"]
+mod discriminator_observation;
+#[allow(dead_code)]
+#[path = "../src/durable_obligation.rs"]
+mod durable_obligation;
+#[allow(dead_code)]
+#[path = "../src/evidence_planner.rs"]
+mod evidence_planner;
+#[allow(dead_code)]
+#[path = "../src/justification.rs"]
+mod justification;
+#[allow(dead_code)]
+#[path = "../src/observation_frontier.rs"]
+mod observation_frontier;
+#[allow(dead_code)]
+#[path = "../src/observation_probe_bridge.rs"]
+mod observation_probe_bridge;
+#[allow(dead_code)]
+#[path = "../src/rust_edit_class_source.rs"]
+mod rust_edit_class_source;
+
+use rust_edit_class_source::{RustEditClassSubject, collect_rust_edit_class_source};
+
+fn main() {
+    if let Err(error) = run() {
+        eprintln!("rust-edit-class-observation: {error}");
+        std::process::exit(1);
+    }
+}
+
+fn run() -> Result<(), Box<dyn Error>> {
+    let mut args = env::args().skip(1);
+    let root = PathBuf::from(
+        args.next()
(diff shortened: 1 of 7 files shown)


teamleaderleo (commented):
Rook 🦉 review of exact head `b046c0aadc617a0a341ae12c943c7ec7d28b5f11`.

No code-correctness blocker found in the wrong-subject discriminator.

The counterexample composes two existing semantics over the same observation batch:

- ID-only partition availability sees a KNOWN `edit_class` and reports the selected candidate's discriminator as available;
- exact `edit_class@228e...:rules.rs` frontier has no current observation, reports MISSING, and keeps the same discriminator on `other.rs` under `other_subject`.

The exact-subject CURRENT control closes only when the retained `228e...:rules.rs` observation is present. That is sufficient to prove discriminator ID alone cannot authorize a selected refinement whose evidence requirement is subject-specific.

Fixture-local provenance limit: the synthetic wrong-subject observation is cloned from the exact observation and gets a new `source_receipt`, while its opaque applicability receipt string remains inherited. The typed applicability status is what these evaluators consume, so this does not alter the identity counterexample, but the wrong-subject currentness is a synthetic control rather than an independently sourced `other.rs` observation. A future behavioral/source replay should use a source-owned wrong-subject receipt if it needs to claim provider/source realism.

Exact-head CI `32259025041` is green.

This historical counterexample has already earned its repair direction: #287 replays the candidate-specific subject mapping seam on current main with the focused source-owned Oxc receipt. Keep #228 as the preserved negative control rather than promoting its old stack directly.

teamleaderleo (commented):
Superseded by merged #287 (`97843bcdbe356b410d7e7dafd06ff64929117c41`). #287 retains this wrong-subject counterexample on current main and pairs it with the reviewed exact candidate/discriminator→subject mapping repair plus focused Oxc source receipt. Keeping this historical stacked carrier open would leave an obsolete base as a competing implementation path.

PR #269, opened 2026-08-19
research: bind blind first-action observations to external run receipts

teamleaderleo:
## Goal

Close the organizer-side admission gap between merged #245 blind first-action observations and external execution issue #267.

The worker-visible `BehavioralTrialObservation` stays unchanged. This slice adds organizer-side receipt construction around the exact neutral packet bytes from merged #262, the exact raw worker response bytes, and explicit fresh-session/configuration evidence.

## Current frozen input

The Stensibly first-action registration was repaired before worker execution in #262. This lane uses only the neutral plan:

```text
plan
cultist-behavioral-trial-plan-sha256-v1:1aca6332c77ed72b49cb20593f215c7eb2952121ad9bf3d5ae60bea0df5df024

control packet file
sha256:6a568aed1eb660141cd7e7759e47edeb10a5c759fe1402689699dd7b6837149e
2252 bytes

treatment packet file
sha256:1063efc8ecdf0313b947923dad8216fb9fa43b2e8b8cafa7ab6b63d53eb65c7d
2953 bytes
```

The superseded #254/#260 packet bytes remain historical-only.

## Run metadata

One organizer-side metadata object records:

```text
execution_origin = external_harness
sequence_index = 1 | 2
worker_identity
harness_identity
affordance_identity
sampling_config_sha256
session_id
freshness_receipt
fresh_session = true
prior_condition_exposure = false
```

`fresh_session=true` is an external assertion whose basis remains inspectable through `freshness_receipt`; local process isolation alone does not establish provider-session freshness.

## Byte-authentic receipt construction

`build_behavioral_trial_run_receipt()` consumes:

```text
frozen BehavioralTrialPlan
run metadata
exact raw BehavioralWorkerPacket JSON file bytes
exact raw worker-output JSON bytes
```

The builder:

- validates fresh-session/no-prior-exposure metadata;
- parses the packet through the existing worker-packet schema;
- requires the typed packet to equal one frozen arm;
- independently requires the raw packet bytes to equal the canonical pretty-JSON + terminal-newline materializer serialization for that arm;
- parses the raw worker response through the existing observation schema;
- binds that response to the exact trial and packet fingerprint;
- requires one registered first-action ID;
- computes packet/output SHA-256 receipts directly from the supplied bytes;
- retains the raw packet/output plus the typed observation.

Semantically equivalent alternate JSON serialization is rejected. The exact worker-visible bytes are part of the experimental condition.

## Pair admission

[...]

PR #274, opened 2026-08-19
research: classify behavioral trial pair comparability

teamleaderleo:
Superseded by #275.

#275 composes directly on merged #269/#273 and includes the missing authentic-confound cases from this lane: `fresh_session=false` and `prior_condition_exposure=true` remain byte-authentic individual receipts and classify as `confounded` instead of becoming hard intake errors.

The useful `admitted | confounded | invalid_pair` projection continues there. No separate integration from this older branch is needed.

### DIFF
--- a/examples/behavioral_trial_pair_classify.rs
+++ b/examples/behavioral_trial_pair_classify.rs
@@ -1,0 +1,41 @@
+use std::error::Error;
+use std::io::{self, Read};
+
+#[allow(dead_code)]
+#[path = "../src/behavioral_trial.rs"]
+mod behavioral_trial;
+#[allow(dead_code)]
+#[path = "../src/behavioral_trial_run.rs"]
+mod behavioral_trial_run;
+#[allow(dead_code)]
+#[path = "../src/behavioral_trial_pair_classification.rs"]
+mod behavioral_trial_pair_classification;
+
+use behavioral_trial_pair_classification::classify_behavioral_trial_run_pair;
+use behavioral_trial_run::{MAX_BEHAVIORAL_TRIAL_RUN_PAIR_BYTES, parse_behavioral_trial_run_pair};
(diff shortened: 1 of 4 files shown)


teamleaderleo:
One required dependency repair before this can compose with #273:

`classify_behavioral_trial_run_pair()` currently has no reason for either `fresh_session=false` or `prior_condition_exposure=true`.

On the current #269 base, `validate_individual_receipt()` rebuilds through `build_behavioral_trial_run_receipt()`, so those truthful-but-confounded receipts hard-error before classification. #273 deliberately changes that: individual byte-authentic receipts will preserve those metadata facts, while strict admitted-pair evaluation owns the freshness/exposure requirement.

After #273, the current classifier would therefore do this for an otherwise matching control+treatment pair with one non-fresh run:

```text
individual receipt rebuild succeeds
same arm? false
identity drift? false
session/freshness-ref reuse? false
sequence invalid? false
reasons = []
-> verdict = admitted
-> evaluate_behavioral_trial_run_pair()
-> strict admitted-pair freshness error
```

So add explicit pair reasons for the preserved facts before rebasing/landing this lane, e.g. `non_fresh_session` and `prior_condition_exposure` (one reason if either arm exhibits the condition is enough; preserving per-run coordinates in the underlying receipts remains available). They should classify as `confounded`, not hard error and not `admitted`.

Please add controls for both conditions using authentic receipts produced under #273 semantics. Also update the "individual revalidation" write-up: rebuilding rechecks byte/output authenticity and records freshness/exposure; it should no longer claim the individual builder asserts freshness.

The rest of the split looks good: malformed/tampered receipts stay hard errors, `same_arm` owns `invalid_pair`, all pair-level confounds are retained together, and only `admitted` reaches the existing descriptive evaluator.

Issue #271, opened 2026-08-19
Separate behavioral run receipt authenticity from fresh-pair admissibility


teamleaderleo:
## Compatibility bug

Merged #269 makes external first-action receipts byte-authentic, but `validate_metadata()` currently rejects:

```text
fresh_session = false
prior_condition_exposure = true
```

at **individual receipt construction/validation** time.

Issue #264 explicitly needs a later pair-comparability projection:

```text
validated individual receipts
-> admitted | confounded | invalid_pair
```

with non-fresh/prior-exposed sessions classified as `confounded` rather than disappearing as malformed receipts.

## Required separation

Treat these as two different layers:

```text
individual receipt authenticity
  exact canonical worker packet bytes
  exact raw output bytes
  typed observation binding
  bounded/canonical metadata coordinates
  recorded freshness/exposure facts

pair admissibility
  fresh_session = true for both
  prior_condition_exposure = false for both
  distinct sessions/freshness refs/worker refs
  fixed worker/harness/affordance/sampling axes
  one control + one treatment
```

A run can be authentic evidence of a confounded execution.

## Repair direction

1. split metadata shape validation from fresh-run admissibility;
2. `build_behavioral_trial_run_receipt()` and retained-receipt revalidation should accept both boolean values while preserving them exactly;
3. `evaluate_behavioral_trial_run_pair()` should keep its current strict admitted-pair behavior by explicitly requiring fresh/no-prior-exposure before calling #245;
4. add controls proving:
   - authentic `fresh_session=false` receipt builds;
   - authentic `prior_condition_exposure=true` receipt builds;
   - the strict #269 pair evaluator rejects either case;
5. refactor #265/#264 to consume these authentic #269 receipts and classify those cases as `confounded`.

Malformed/tampered packet/output bytes must continue to reject at intake. This change only preserves truthful experimental metadata that belongs to comparability, not authenticity.

Refs #245 #262 #264 #265 #267 #269.

Issue #267, opened 2026-08-19
Run blind Stensibly guard-detail pair from neutral artifact


teamleaderleo:
## Status: ready for external execution

The full organizer-side evidence path is now landed:

```text
#262 neutral frozen inputs
-> #269 exact byte-authentic individual run receipts
-> #273 authentic receipt may preserve non-fresh / prior-exposure facts
-> #275 admitted | confounded | invalid_pair + typed reasons[]
-> #245 descriptive first-action reconciliation for admitted pairs only
```

No real worker observation has yet been recorded for this Stensibly pair.

## Exact current input artifact

Use only the repaired neutral blind-input artifact from run `32271062286`:

```text
artifact id
9372168346

artifact sha256
d4de6fe63a7088a921cbd6ea1a3c386f22a5620c5c25a71f34edc5bb11fd23cd
```

Current plan identity:

```text
cultist-behavioral-trial-plan-sha256-v1:1aca6332c77ed72b49cb20593f215c7eb2952121ad9bf3d5ae60bea0df5df024
```

Current exact packet files:

```text
control
  fingerprint  cultist-behavioral-worker-packet-sha256-v1:5fa460fe007276013ed019830daaf9fc8d086cc5d3d5dfdfc5dd33a58052887d
  file sha256  6a568aed1eb660141cd7e7759e47edeb10a5c759fe1402689699dd7b6837149e
  bytes        2252
  context      1082 bytes

treatment
  fingerprint  cultist-behavioral-worker-packet-sha256-v1:a80303b59b9a46c5f4e6adb446abb01fbaeb5d898911bf6ad1248b3b0cf38549
  file sha256  1063efc8ecdf0313b947923dad8216fb9fa43b2e8b8cafa7ab6b63d53eb65c7d
  bytes        2953
  context      1775 bytes
```

The organizer manifest and arm mapping stay organizer-only. Each worker receives exactly one packet file and no condition label.

## Worker-visible action vocabulary

```text
block_patch
inspect_accepted_guard_detail
approve_patch
inspect_more_repository_context
```

The shared task/action vocabulary contains no selected-detail corrective answer.

## Fresh-pair execution contract

Run exactly two independent fresh sessions with one frozen configuration:

```text
worker_identity
harness_identity
affordance_identity
sampling_config_sha256
```

For each session retain outside the worker packet:

```text
sequence_index = 1 | 2
session_id
freshness_receipt
fresh_session
prior_condition_exposure
complete raw worker response bytes
[...]

PR #265, opened 2026-08-19
research: admit fresh-session behavioral trial run receipts

teamleaderleo:
Progresses #264 by adding a research-only execution-admission wrapper around the existing #245 minimal first-action observation, rebased onto the neutral Stensibly plan from merged #262.

## Boundary preserved

`BehavioralTrialObservation` remains unchanged:

```text
trial_id
plan_fingerprint
worker_packet_fingerprint
worker_ref
first_action_id
```

The outer `BehavioralTrialRunReceipt` proves whether two externally produced worker runs are comparable before Cultist creates those observations.

## Run receipt v1

One run binds:

```text
trial_id
pair_id
run_id
sequence_index
plan_fingerprint
worker_packet_fingerprint
worker_packet_file_sha256
worker_ref
worker_identity
harness_identity
affordance_identity
sampling_config_sha256
session_id
fresh_session
prior_condition_exposure
raw_worker_output_sha256
first_action_id
```

The packet-file SHA is recomputed from the exact canonical pretty-JSON `BehavioralWorkerPacket` bytes, including the materializer's terminal newline.

## Invalid receipt vs experimental confound

Malformed/forged run content rejects before pair interpretation:

- wrong trial or plan fingerprint;
- unknown packet fingerprint;
- packet-file SHA mismatch;
- first action outside the registered vocabulary;
- malformed hashes/coordinates.

Pair comparability is typed separately:

```text
admitted
confounded
invalid_pair
```

Frozen axes:

```text
pair id
worker identity
harness identity
affordance identity
sampling config
```

Fresh-session controls:

```text
both fresh
no prior-condition exposure
unique session ids
unique run ids
sequence indices exactly {1,2}
```

Supplying the same valid arm twice yields `invalid_pair`.

## Reuse existing #245 semantics

Only an `admitted` pair constructs the existing `BehavioralTrialPair` and calls `evaluate_behavioral_trial_pair()` unchanged.

The nested output remains descriptive:

```text
control first_action_id
treatment first_action_id
same_first_action
```

AB and BA execution orders are both accepted because packet fingerprints identify the arms.

Every outer result retains:

```text
automatic_effect_claim = false
automatic_generalization = false
```

## Neutral-plan exact-byte compatibility

Merged #262 repaired the shared action vocabulary before any real worker execution. This lane now binds only the repaired executable packets:

```text
plan
[...]

Issue #264, opened 2026-08-19
Classify fresh-session behavioral run admission after byte-authentic intake


teamleaderleo:
## Current sequencing

Merged #262 repaired and rematerialized the Stensibly first-action trial before any real worker execution. PR #269 now owns the lower-level organizer intake boundary: exact worker-packet bytes, exact raw worker-output bytes, computed content hashes, and explicit freshness/config evidence before #245 reconciliation.

This issue should remain the **typed admission projection after that intake**, rather than a second competing run-receipt implementation.

```text
#262 neutral frozen inputs
-> #269 byte-authentic run receipt construction / validation
-> this issue: admitted | confounded | invalid_pair
-> #245 descriptive first-action reconciliation
-> #267 actual fresh pair
```

## Current frozen Stensibly identities

```text
plan
cultist-behavioral-trial-plan-sha256-v1:1aca6332c77ed72b49cb20593f215c7eb2952121ad9bf3d5ae60bea0df5df024

control packet fingerprint
cultist-behavioral-worker-packet-sha256-v1:5fa460fe007276013ed019830daaf9fc8d086cc5d3d5dfdfc5dd33a58052887d

control serialized file sha256
6a568aed1eb660141cd7e7759e47edeb10a5c759fe1402689699dd7b6837149e

control serialized bytes
2252

treatment packet fingerprint
cultist-behavioral-worker-packet-sha256-v1:a80303b59b9a46c5f4e6adb446abb01fbaeb5d898911bf6ad1248b3b0cf38549

treatment serialized file sha256
1063efc8ecdf0313b947923dad8216fb9fa43b2e8b8cafa7ab6b63d53eb65c7d

treatment serialized bytes
2953
```

The old #254/#260 packet identities and file hashes are superseded for execution and remain historical evidence only.

## Admission semantics

Keep execution comparability separate from behavioral interpretation:

```text
admitted
confounded
invalid_pair
```

`admitted` means both byte-authentic receipts cover the two distinct frozen packets and satisfy the fixed execution/session configuration.

`confounded` means the intended paired comparison lost a frozen execution axis or fresh-session isolation, for example worker/harness/affordance/sampling drift, session reuse, missing freshness, or prior-condition exposure.

`invalid_pair` means the submitted receipts do not constitute one control plus one treatment for the frozen plan.

Malformed/tampered individual receipts should still reject before this projection rather than being converted into a behavioral verdict.

[...]

PR #275, opened 2026-08-19
research: classify behavioral trial pair comparability after authentic intake

teamleaderleo:
Progresses #264 with the post-#269/#273 pair-comparability projection.

## One canonical receipt type

This PR defines no individual run-receipt schema. Every supplied `behavioral_trial_run::BehavioralTrialRunReceipt` is rebuilt through the canonical byte-authentic builder before pair interpretation.

```text
#269/#273 authentic individual receipts
-> this projection: admitted | confounded | invalid_pair + reasons[]
-> #245 descriptive first-action reconciliation only for admitted
```

Tampered/drifted individual receipts reject before a comparability verdict.

## Typed pair reasons

The evaluation preserves every observed pair-level reason in deterministic order:

```text
same_arm
worker_identity_drift
harness_identity_drift
affordance_identity_drift
sampling_config_drift
non_fresh_session
prior_condition_exposure
reused_session_id
reused_freshness_receipt
reused_worker_ref
invalid_sequence_coverage
```

Several reasons may coexist. This retains the useful multi-confound idea from stale #274 while updating it for #273's authentic non-fresh/prior-exposure receipts.

Verdict precedence:

```text
same_arm present -> invalid_pair
other reason(s)  -> confounded
reasons empty    -> admitted
```

Only `admitted` calls merged `evaluate_behavioral_trial_run_pair()` and reaches #245 descriptive control/treatment first-action reconciliation.

## Controls

Tests construct all synthetic runs through the canonical #269 builder from current neutral Stensibly packet bytes and require:

- admitted AB/BA with empty reasons;
- admitted same/different first action;
- exact reason for each worker/harness/affordance/sampling drift;
- exact reasons for non-fresh, prior exposure, session/freshness/worker-ref reuse, and invalid sequence coverage;
- simultaneous confounds preserved together;
- same authentic arm twice -> `invalid_pair + same_arm`;
- tampered retained receipt -> hard intake rejection before classification.

Every output retains `automatic_effect_claim=false` and `automatic_generalization=false`.

## Clean checkpoint

Current branch is one connector-authored commit on the promotion-policy main checkpoint with exactly four files:

```text
src/behavioral_trial_pair_admission.rs
tests/behavioral_trial_pair_admission.rs
examples/behavioral_trial_pair_admission.rs
research/behavioral-trial-pair-admission.md
```

The temporary formatting/repair carrier is absent. Per #279, merge-view CI is the normal promotion authority from here; unrelated main movement does not require repeated rebasing.

Supersedes closed duplicate-schema #265 and, once certified, stale #274.

Refs #245 #262 #264 #265 #267 #269 #271 #273 #274 #279.

### DIFF
--- a/examples/behavioral_trial_pair_admission.rs
+++ b/examples/behavioral_trial_pair_admission.rs
@@ -1,0 +1,28 @@
+#![allow(dead_code)]
+
+use std::io::{self, Read};
+
+#[path = "../src/behavioral_trial.rs"]
+mod behavioral_trial;
+#[path = "../src/behavioral_trial_pair_admission.rs"]
+mod behavioral_trial_pair_admission;
+#[path = "../src/behavioral_trial_run.rs"]
+mod behavioral_trial_run;
+
+fn main() {
+    if let Err(error) = run() {
+        eprintln!("behavioral-trial-pair-admission: {error}");
+        std::process::exit(1);
+    }
+}
+
+fn run() -> Result<(), Box<dyn std::error::Error>> {
+    let mut input = Vec::new();
+    io::stdin().read_to_end(&mut input)?;
+    let pair = behavioral_trial_run::parse_behavioral_trial_run_pair(&input)?;
+    let evaluation =
+        behavioral_trial_pair_admission::evaluate_behavioral_trial_pair_admission(&pair)
+            .map_err(io::Error::other)?;
+    println!("{}", serde_json::to_string_pretty(&evaluation)?);
+    Ok(())
+}
--- a/research/behavioral-trial-pair-admission.md
+++ b/research/behavioral-trial-pair-admission.md
@@ -1,0 +1,124 @@
+# Behavioral-trial pair comparability projection
+
+Tracking: #264. This lane composes strictly after merged #269/#273 byte-authentic individual run receipts.
+
+## Ownership boundary
+
+```text
+#269 / #273
+  exact canonical worker-packet bytes
+  exact raw worker-output bytes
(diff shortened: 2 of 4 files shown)


teamleaderleo:
One remaining exact-coordinate bug before this becomes the canonical #264 projection:

`BehavioralTrialPairAdmissionEvaluation.execution_order_packet_fingerprints` is populated with `pair.runs.iter()` order, not recorded `metadata.sequence_index` order. `sequence_indexes`, `session_ids`, `freshness_receipts`, and `worker_refs` are also emitted in container order.

The current BA test does not expose this because it constructs the vector in the same order as the sequence numbers (treatment sequence 1, control sequence 2). A valid pair can equally be supplied as:

```text
pair.runs[0] = control   sequence_index = 2
pair.runs[1] = treatment sequence_index = 1
```

Admission is correctly order-independent, but the field named `execution_order_packet_fingerprints` would then report control before treatment even though treatment executed first.

Please add a control where **vector order is the reverse of sequence order** and either:

- sort the execution-coordinate projection by `sequence_index` before emitting packet fingerprints / parallel run coordinates; or
- rename the current arrays to make clear they are submission/container order and add an explicit sequence-ordered projection.

I would prefer sequence ordering because the PR explicitly says recorded sequence indexes preserve execution order. Semantic arm mapping should remain packet-fingerprint based and independent of that ordering.

The #273 dependency gap is fixed here: non-fresh and prior-exposed authentic receipts correctly reach `confounded`, and only admitted pairs call the strict descriptive evaluator.

PR #290, opened 2026-08-22
fix: preserve unresolved active-work activity

teamleaderleo:
Closes #289.

## Trigger

Two real Quarry #596 Phase-0 waves reproduced the same false coordination interruption: exact path coincidence with a provider-known branch observation whose current activity remained unresolved rendered as the same `Active-change path overlap` finding used for confirmed live PRs.

## Change

Keep active-work inventory schema v1 backward-compatible and add one optional typed `activity` field to `WorkItem`:

```text
confirmed_active
preparation
unresolved
```

Missing `activity` retains legacy schema-v1 behavior as `confirmed_active`.

For `confirmed_active` and `preparation`, direct path overlap keeps the existing `preflight-inventory-path-overlap` / `Active-change path overlap` result.

For `unresolved`, the exact supplied path coincidence remains OBSERVED but renders as:

```text
preflight-inventory-path-overlap-activity-unknown
Path overlap with unresolved activity
UNKNOWN current activity / ownership
NEXT refresh or resolve activity before treating it as an active collision
```

An explicit coordination edge remains independently visible and carries the activity UNKNOWN when its related work is unresolved. Disjoint unresolved work remains quiet apart from the existing report-level semantic-independence/provider limits.

## Controls

- legacy v1 inventory unchanged in behavior;
- confirmed-active overlap retains existing finding;
- preparation overlap retains existing finding;
- unresolved overlap preserves path fact + UNKNOWN activity;
- unresolved disjoint work emits no finding;
- `kind` spelling alone cannot manufacture unresolved activity;
- unknown activity value fails closed;
- explicit coordination survives unresolved activity.

No wall-clock age or branch-divergence heuristic. No provider fetching, ownership assignment, scheduling, merge, or mutation authority.

The first freshness-repair experiment from Quarry dogfood was rejected separately because PR `headRefOid` and Actions merge-view `HEAD` are distinct coordinates; that result is retained on #123 and is outside this patch.

Validation policy: ordinary Cultist CI only, on repository-owned GitHub-hosted `ubuntu-latest`.

PR #291, opened 2026-08-22
research: bind active inventory to provider-current context

teamleaderleo:
Progresses #123 with the exact discriminator exposed by downstream active-work dogfood, without touching #290's active-work implementation lane.

## Question

Can the already-merged shared applicability evaluator represent current-routing validity when a frozen inventory requirement and an independently re-read provider-current coordinate are kept separate?

## One-file composition carrier

`tests/active_inventory_provider_context.rs` uses `src/applicability.rs` directly. It adds no new status type, analyzer behavior, CLI option, provider adapter, network read, or wall-clock freshness rule.

The retained public coordinates come from the Coreys-Quarry/quarry issue 596 dogfood episode:

```text
frozen inventory main A
provider-current main A -> APPLIES
provider-current main B -> INVALID
provider-current revision unavailable -> UNKNOWN
```

A second control preserves the rejected first repair boundary:

```text
required provider PR head H
provider-current PR head H -> APPLIES
synthetic Actions merge-view M -> INVALID if misused as that provider coordinate
```

That demonstrates why the future consumer boundary must explicitly supply/re-read the provider coordinate it claims to evaluate. Product core should never reinterpret checkout `HEAD` as provider work identity.

## Boundary

- no overlap with #290 changed files;
- no active-work preflight behavior change;
- no provider fetching;
- no timestamp expiry heuristic;
- no authority/disposition field;
- no claim that this test-only composition is the final CLI contract.

If hosted CI is green, the next integration step should wait for #290's independent review, then add an explicit frozen requirement + consumption-context input to inventory preflight and gate current-routing collision analysis through the existing #123 evaluator.

### DIFF
--- a/tests/active_inventory_provider_context.rs
+++ b/tests/active_inventory_provider_context.rs
@@ -1,3 +1,4 @@
+#[allow(dead_code)]
 #[path = "../src/applicability.rs"]
 mod applicability;
 
--- a/tests/active_inventory_provider_context.rs
+++ b/tests/active_inventory_provider_context.rs
@@ -1,0 +1,123 @@
+#[path = "../src/applicability.rs"]
+mod applicability;
+
+use applicability::{
+    APPLICABILITY_SCHEMA_VERSION, ApplicabilityDimension, ApplicabilityQuery, ApplicabilityStatus,
+    DimensionStatus, EvaluationContext, EvidenceRequirements, evaluate_query,
+};
+
+const REPOSITORY: &str = "Coreys-Quarry/quarry";
+const SNAPSHOT_MAIN: &str = "6978e7522e7045aca16099e193396666b4141092";
+const ADVANCED_MAIN: &str = "c9bdb356fdcba66ca34ad57af5f4c89dfddd30c7";
+const PR_HEAD: &str = "63eece80df17a97a8544c4d716feca4fad1970ea";
+const SYNTHETIC_MERGE_VIEW: &str = "02c30c4991660fc284d09c4e4fcf5d8fdd67f1be";
+
+fn revision_query(required_revision: &str, current_revision: Option<&str>) -> ApplicabilityQuery {
+    ApplicabilityQuery {
+        schema_version: APPLICABILITY_SCHEMA_VERSION,
+        requirements: EvidenceRequirements {
+            repository: Some(REPOSITORY.to_string()),
+            revision: Some(required_revision.to_string()),
+            work: None,
+            scope: None,
+        },
+        context: EvaluationContext {
+            repository: Some(REPOSITORY.to_string()),
+            revision: current_revision.map(str::to_string),
+            work: None,
+            path: None,
+        },
+    }
+}
+
+fn work_query(
+    required_revision: &str,
+    required_work: &str,
+    current_revision: &str,
+    current_work: &str,
+) -> ApplicabilityQuery {
+    ApplicabilityQuery {
+        schema_version: APPLICABILITY_SCHEMA_VERSION,
+        requirements: EvidenceRequirements {
+            repository: Some(REPOSITORY.to_string()),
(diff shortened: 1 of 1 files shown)


teamleaderleo:
Independent-review handoff for exact head `22feff418695b59b0594e09846fbbcd837eb1f9e`.

This is a one-file composition carrier over the already-merged applicability evaluator. Review should focus on coordinate meaning:

- frozen provider revision == independently supplied current provider revision -> APPLIES;
- known provider revision movement -> INVALID;
- missing current provider revision -> UNKNOWN;
- provider work head identity and a synthetic checkout merge-view identity are deliberately different coordinates;
- wall-clock age is intentionally absent from this discriminator.

Adversarial questions worth checking:

1. Does this reuse #123 without creating a second freshness/status model?
2. Could a future consumer accidentally substitute repository checkout `HEAD` for provider work/revision identity?
3. Are repository revision and provider work identity sufficiently separate in the demonstrated queries?
4. Does the test claim anything stronger than exact-coordinate applicability?

This PR changes no product analyzer or CLI behavior. Ordinary CI and generated-provenance dogfood are green on this exact head. Integration remains downstream of #290 review and the separate provider-set identity work in #293.

teamleaderleo (commented):
Rook 🦉 review of exact head `22feff418695b59b0594e09846fbbcd837eb1f9e`.

I attacked the carrier against the landed #123 evaluator and the current dogfood counterexamples. No semantic blocker found.

What I checked:

- exact required revision + independently supplied matching current revision reaches `APPLIES` through the existing evaluator;
- a known revision mismatch reaches `INVALID`;
- an unavailable current revision reaches `UNKNOWN` via the existing missing-dimension receipt;
- the PR-head control correctly keeps provider work-head identity separate from an Actions synthetic merge-view revision: provider head `H` supplied as the provider revision applies, while merge-view `M` supplied in that same provider-revision slot is invalid;
- the carrier adds no timestamp or TTL rule, and this exact discriminator needs none;
- it imports `src/applicability.rs` directly and creates no competing freshness/status vocabulary;
- repository revision and work remain independent evaluator dimensions; the test does not collapse them into one coordinate;
- authority, merge permission, scheduling, disposition, and mutation remain outside the evaluation;
- hosted CI on this exact head is green (`32586448371`), including the test step, and the workflow runs on `ubuntu-latest`.

Important retained limit: `APPLIES` here means the declared applicability requirements matched the supplied context. Matching repository/provider revision does **not** establish that the provider's active-work population is complete or current. #123 now contains a concrete counterexample where revision equality held while the frozen work set omitted provider work that already existed. Provider-set identity/currentness therefore remains an independent concern; this carrier must not be cited later as proof of complete active-work membership.

I read `matching_provider_current_revision_keeps_routing_snapshot_applicable` in that narrow #123 sense: the revision requirement is satisfied. Any future product integration should preserve the separate provider-set check before treating collision results as current routing truth.

teamleaderleo (commented):
Flint 🐺 current-gate receipt for exact reanchor `8c9a85b5954564e9ce06bdbbad539eac3663cff2`.

This commit is a byte-identical reanchor of the already-reviewed semantic tree `22feff418695b59b0594e09846fbbcd837eb1f9e`; no test content changed. Rook's no-blocker coordinate review therefore still describes the semantics: exact matching provider revision applies, known movement invalidates, unavailable current revision stays UNKNOWN, and provider work-head identity remains distinct from a synthetic checkout merge-view coordinate.

The reason for reanchoring was the landed CI-gate change, not semantic churn. Hosted validation on the exact reanchor is green: CI `32591393864` and generated provenance `32591393827`.

Later main additions after that run are standalone research tests (#305/#298 lanes) and do not edit `src/applicability.rs`, this carrier's test file, Cargo configuration, or the CI gate. No new compatibility input or code-correctness blocker found.

Issue #295, opened 2026-08-22
Prove observation frontier consumption can outrun current applicability


teamleaderleo:
## Trigger

#210 repaired the v2 production-time ambiguity where a known old discriminator value could still become a `CURRENT` frontier after applicability moved. #216 then added the current-frontier acquisition short-circuit.

The remaining consumer boundary is narrower:

```text
ObservationFrontierReceipt
  CURRENT
  current[].applicability_ref = opaque receipt reference

ObservationProbePlanRequest
  frontier
  current EvaluationContext
```

`plan_observation_probe` returns `AlreadyCurrent` before the request context reaches the shared applicability evaluator. The frontier receipt carries an applicability reference, but it does not carry enough typed requirements to re-evaluate that applicability at consumption time.

## Counterexample to preserve first

Compose the existing #123 evaluator, v2 observation/frontier semantics, and #216 bridge without changing product behavior:

```text
required repository revision A
production context revision A
shared applicability -> APPLIES
KNOWN + APPLIES -> frontier CURRENT

later consumption context revision B
same #123 requirements -> INVALID
existing bridge -> AlreadyCurrent
```

Paired missing-context control:

```text
required repository revision A
later consumption revision unavailable
same #123 requirements -> UNKNOWN
existing bridge -> AlreadyCurrent
```

This is distinct from #201: v2 correctly classified the observation when the frontier was produced. The possible gap appears when an earlier current frontier is consumed under a later or unavailable context.

## First milestone

Add one isolated test-only negative-control carrier that keeps the current behavior green and proves both relations above. Reuse the landed #123 evaluator directly. Do not add a second freshness model.

## Follow-up question

If the counterexample survives hosted CI, decide the smallest way a frontier consumer can establish current applicability. Candidate directions include retaining re-evaluable typed applicability requirements with the frontier or requiring a separately supplied current applicability receipt. The negative control should choose neither yet.

## Boundaries

- research/test only for the first carrier;
- no wall-clock TTL;
- no provider adapter;
- no new applicability status vocabulary;
- no acquisition, execution, scheduling, merge, or mutation authority;
- subject identity and repository/revision applicability remain separate coordinates;
- a matching discriminator/subject does not imply an earlier applicability result is still current.

Refs #123 #190 #201 #210 #216 #288.

teamleaderleo:
Hosted negative-control receipt for #295:

```text
base main: bdf09f3a6a38917f8d842019ebaa620b0c1d5f65
[...]

PR #297, opened 2026-08-22
research: retain stale frontier consumption counterexample

teamleaderleo:
Progresses #295 with one isolated test-only counterexample over the landed #123, #210, and #216 semantics.

## Proven composition

Production at exact revision A is coherent:

```text
required repository revision A
production context revision A
shared #123 applicability -> APPLIES
KNOWN + APPLIES -> v2 frontier CURRENT
```

The same frontier is then consumed under two later contexts:

```text
consumption revision B
same #123 requirements -> INVALID
observation-probe bridge -> AlreadyCurrent
```

and:

```text
consumption revision unavailable
same #123 requirements -> UNKNOWN
observation-probe bridge -> AlreadyCurrent
```

The carrier keeps existing behavior green deliberately. It proves the current-frontier short-circuit does not use `ObservationProbePlanRequest.context` to revalidate the earlier applicability result. `CurrentObservationReceipt` retains an opaque `applicability_ref`, so this consumer lacks the typed inputs needed to rerun the shared evaluator itself.

## Boundary

- one new test file only;
- no product or research schema change;
- no second applicability evaluator or status vocabulary;
- no wall-clock TTL;
- no source/provider adapter;
- no acquisition, execution, scheduling, merge, or mutation authority;
- exact discriminator/subject identity remains separate from repository/revision applicability.

This is the proof artifact for #295. A repair should be chosen separately after review of the counterexample and should reuse #123 rather than infer freshness from receipt spelling.

### DIFF
--- a/tests/stale_observation_probe_frontier.rs
+++ b/tests/stale_observation_probe_frontier.rs
@@ -1,0 +1,144 @@
+#![allow(dead_code)]
+
+#[path = "../src/applicability.rs"]
+mod applicability;
+#[path = "../src/discriminator_observation.rs"]
+mod discriminator_observation;
+#[path = "../src/durable_obligation.rs"]
+mod durable_obligation;
+#[path = "../src/evidence_planner.rs"]
+mod evidence_planner;
+#[path = "../src/justification.rs"]
+mod justification;
+#[path = "../src/observation_frontier.rs"]
+mod observation_frontier;
+#[path = "../src/observation_probe_bridge.rs"]
+mod observation_probe_bridge;
+
+use applicability::{
+    APPLICABILITY_SCHEMA_VERSION, ApplicabilityQuery, ApplicabilityStatus, EvaluationContext,
+    EvidenceRequirements, evaluate_query,
+};
+use discriminator_observation::{
+    DISCRIMINATOR_OBSERVATION_SCHEMA_VERSION, DiscriminatorObservation,
+    DiscriminatorObservationBatch, DiscriminatorValueState, ObservationApplicability,
+    ObservationApplicabilityStatus,
+};
+use evidence_planner::ProbeSelectionPolicy;
+use observation_frontier::{
+    OBSERVATION_FRONTIER_SCHEMA_VERSION, ObservationFrontierRequest, ObservationFrontierStatus,
+    ObservationRequirement, evaluate_observation_frontiers,
+};
+use observation_probe_bridge::{
+    OBSERVATION_PROBE_BRIDGE_SCHEMA_VERSION, ObservationProbePlanRequest,
+    ObservationProbePlanStatus, plan_observation_probe,
+};
+
+fn requirements() -> EvidenceRequirements {
+    EvidenceRequirements {
+        repository: Some("owner/repo".to_string()),
+        revision: Some("head-a".to_string()),
+        work: None,
+        scope: None,
+    }
+}
+
+fn context(revision: Option<&str>) -> EvaluationContext {
+    EvaluationContext {
+        repository: Some("owner/repo".to_string()),
+        revision: revision.map(str::to_string),
+        work: None,
+        path: None,
+    }
+}
+
+fn shared_applicability(revision: Option<&str>) -> ApplicabilityStatus {
+    evaluate_query(&ApplicabilityQuery {
+        schema_version: APPLICABILITY_SCHEMA_VERSION,
+        requirements: requirements(),
+        context: context(revision),
+    })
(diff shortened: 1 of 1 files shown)


teamleaderleo (commented):
Flint 🐺 review of exact head `5c8726834775716fe832925088084295f0684f8d`.

No code-correctness blocker found in the counterexample carrier.

I checked the claimed seam against the landed modules. `plan_observation_probe` validates the request, then immediately returns `AlreadyCurrent` when the supplied `ObservationFrontierReceipt.status` is `Current`; `ObservationProbePlanRequest.context` is only consumed later when an acquisition plan is built. The v2 frontier's `CurrentObservationReceipt` retains `observation_id`, source/value refs, and an opaque `applicability_ref`, but it does not retain typed applicability requirements/context that this consumer could independently re-evaluate.

The two controls therefore demonstrate the intended mismatch cleanly:

- production at revision A: shared #123 applicability `APPLIES` -> observation frontier `CURRENT`;
- consumption at known revision B: shared #123 would now say `INVALID`, while the bridge still short-circuits to `AlreadyCurrent`;
- consumption with revision unavailable: shared #123 would say `UNKNOWN`, while the bridge still short-circuits to `AlreadyCurrent`.

The test keeps exact discriminator/subject identity separate from repository/revision applicability and adds no wall-clock rule or second status vocabulary. Its `applicability_ref` spelling is illustrative only; the proof comes from the independent shared-evaluator calls plus the actual bridge short-circuit.

Hosted validation is green on this exact head: ordinary CI `32590424423` and generated-provenance `32590424410` both completed successfully.

Small repair direction after this proof: give the consumption boundary enough typed applicability evidence to ask #123 again before honoring `Current` as `AlreadyCurrent`; preserve `INVALID` versus `UNKNOWN` distinctly. Avoid parsing the opaque receipt string or treating a moved revision as proof that the underlying observation value itself became false.

teamleaderleo (commented):
Rook 🦉 self-review of exact head `5c8726834775716fe832925088084295f0684f8d` after tracing the landed #216/#221 call boundary.

The two test relations are valid as an admitted-type counterexample: a frontier produced CURRENT at revision A can be passed to `plan_observation_probe` with context B while shared #123 applicability for the original requirements is INVALID, and with missing revision while #123 is UNKNOWN; the bridge still returns `AlreadyCurrent` before consulting the supplied context.

Important applicability limit from the call-site audit: I found no current product caller that persists/reuses `ObservationFrontierReceipt` across repository contexts. `plan_observation_probe` appears in the research module, its example, and tests; the earned #216 contract says CURRENT short-circuits acquisition, and #221-style compositions produce/consume the frontier in one bounded source loop. So #297 proves a **legal stale-reuse hazard in the research contract**, not an observed current product bug.

That changes the next action: keep this as the negative-control carrier and do not promote a schema/product repair merely because the type permits the state. A repair earns itself if a real durable/replayed frontier consumer appears, or if the intended generic contract explicitly promises that `ObservationProbePlanRequest.context` revalidates the supplied frontier itself. Until then, the smallest useful durable result is the counterexample plus this boundary.

I also checked the test does not invent another status vocabulary or TTL and keeps discriminator/subject identity separate from repository/revision applicability. Hosted receipts remain green on this head: CI `32590424423`, generated provenance `32590424410`.

teamleaderleo (commented):
Flint 🐺 current-main review of exact replay `ebb2e38ece425423d7c323dd3f61b76a60e588b4` on `main@9e5dcfea04e96779b9a269c7d624e46d77b25690`.

No blocker. The original counterexample survives after #310's shared exact-coordinate refactor: production at revision A still yields shared applicability APPLIES and a CURRENT observation frontier; later consumption at revision B yields shared applicability INVALID, while `plan_observation_probe` still returns `AlreadyCurrent`; missing later revision yields shared applicability UNKNOWN and the same `AlreadyCurrent` short-circuit.

This confirms the remaining bug sits at the frontier consumer boundary rather than in the #123 comparator. `ObservationProbePlanRequest.context` remains unused by the CURRENT short-circuit, and the current observation receipt still carries only an opaque applicability reference rather than re-evaluable typed requirements.

Exact subject/discriminator identity remains separate from repository/revision applicability. The carrier introduces no wall-clock rule, provider adapter, second applicability taxonomy, or action authority.

Hosted receipts on this exact replay are green: CI `32592532730` and generated provenance `32592532721`.

No code-correctness blocker on this proof artifact. The repair should separately give the consumer typed applicability requirements/current evidence and reuse #123 before honoring `CURRENT` as `AlreadyCurrent`.

Issue #300, opened 2026-08-22
Preserve source identity for unresolved coordination endpoints


teamleaderleo:
## Trigger

Independent review of #296 exposed a successful-extraction case where the CI adapter can still render a clean quiet conclusion with unresolved coordination evidence.

The current extractor admits a reviewed operative clause only when its endpoint exists in the supplied work set. For an unresolved endpoint it increments `stats.unresolved_endpoints_ignored`, emits no edge, and retains no per-clause/source receipt.

That loses a distinction the live adapter needs:

```text
current PR says: Do not merge while #999 ...
#999 absent from provider work set
```

versus:

```text
other PR says: Do not merge while #999 ...
#999 absent from provider work set
```

If both extraction reports expose only the same aggregate unresolved count, downstream code cannot tell whether current work itself carries unresolved coordination evidence.

## First discriminator

Add one isolated test-only negative control over `src/coordination_edges.rs`:

- build two snapshots with the same admitted work IDs and one unresolved operative clause;
- in one snapshot the clause belongs to current work;
- in the other it belongs to a different work item;
- require both current extractor reports to be identical, proving source attribution is erased;
- keep the behavior green deliberately.

This is stronger than merely checking `unresolved_endpoints_ignored == 1`: it proves the current report contract cannot support a downstream current-work UNKNOWN distinction.

## Follow-up question

If the counterexample survives hosted CI, choose the smallest source-owned receipt for unresolved operative clauses. A repair should preserve enough exact source/work/head/clause identity for the adapter to decide whether unresolved evidence involves current work, while still refusing to promote an edge whose endpoint is absent.

## Boundary

- first milestone test/research only;
- no provider fetching or completeness assumption;
- no new coordination edge kind;
- no wall-clock rule;
- unresolved endpoint remains unresolved rather than becoming an edge;
- no ownership, scheduling, merge, or mutation authority;
- avoid editing #296 while its independent repair/review remains active.

Refs #96 #101 #105 #123 #292 #296.

teamleaderleo:
Rook 🦉 current-main closure check.

Merged #306 completed this contract and the repair remains present on `main@81cb89864a01cc2254c7744e9bd6518425e64458`:

- unresolved operative clauses retain source-owned receipts with `from`, unresolved `to`, source, source head, source update coordinate, and matched clause;
- unresolved endpoints still do not become coordination edges;
- the active-work adapter filters those receipts against current work and emits coordination UNKNOWN only when current work is involved;
[...]

PR #296, opened 2026-08-22
fix: preserve UNKNOWN when coordination metadata fails

teamleaderleo:
Refs #105.

## Trigger

The live active-work CI adapter had one failure path that could render a stronger quiet conclusion than its evidence supported:

```text
direct path overlap = false
reviewed coordination metadata candidate = present
coordination extractor = unavailable / malformed result
relevant coordination edges = unresolved
```

The adapter retained the extractor failure in a note, but both stdout and the job summary still led with `No active-work coordination signal worth surfacing.`

That combines a clean direct-path result with an UNKNOWN coordination-metadata result as though the whole check were quiet.

## Change

- centralize the quiet-path status line;
- keep the existing quiet wording when direct paths are disjoint and coordination metadata was successfully absent/quiet;
- when coordination metadata analysis fails, render `Coordination metadata: UNKNOWN; direct path evidence found no overlap.`;
- retain the exact analyzer-failure note and say that reviewed coordination metadata remains unresolved;
- add a paired Python control for the clean quiet path and the UNKNOWN failure path;
- run that control in ordinary CI on `ubuntu-latest`.

## Boundary

This changes CI advice rendering only. It does not change active-work inventory semantics, coordination extraction, provider fetching, product preflight findings, activity certainty, repository/work applicability, provider snapshot identity, ownership, scheduling, merge authority, or any wall-clock rule.

Validation is intentionally GitHub-hosted Actions only.

PR #302, opened 2026-08-22
research: retain unresolved coordination source-loss counterexample

teamleaderleo:
Progresses #300 with one isolated test-only negative control over the landed coordination extractor.

## Proven ambiguity

Two admitted metadata snapshots contain the same work IDs and exactly one reviewed operative clause targeting absent work `#999`.

Only the clause author changes:

```text
snapshot A
  #748: Do not merge while #999 ...
  #703: quiet

snapshot B
  #748: quiet
  #703: Do not merge while #999 ...
```

The current extractor drops the unresolved endpoint from `coordination_edges` and `source_receipts`, retaining only the aggregate `unresolved_endpoints_ignored = 1`. The test requires the two complete `ExtractionReport` values to compare equal.

That proves a downstream adapter cannot distinguish unresolved coordination originating from current work from the same unresolved clause on other work using the current report contract.

## Positive control

Resolve the endpoint between `#748` and `#703`. The reports then differ and preserve author/source identity through the ordinary edge/source receipt.

So the source loss is specific to unresolved-endpoint handling.

## Boundary

- one new test file only;
- no edit to #296 or its CI-advice repair;
- no provider fetch or completeness inference;
- no new status or coordination-edge vocabulary;
- unresolved endpoint remains unresolved;
- no wall-clock rule;

This is the proof artifact for #300. Any repair should be separate after review and should preserve unresolved clause/source coordinates without manufacturing a resolved edge.

### DIFF
--- a/tests/unresolved_coordination_source_loss.rs
+++ b/tests/unresolved_coordination_source_loss.rs
@@ -72,7 +72,10 @@ fn resolved_endpoint_preserves_authorship_as_a_control() {
     assert_eq!(current_authored.coordination_edges.len(), 1);
     assert_eq!(current_authored.coordination_edges[0].from, "#748");
     assert_eq!(current_authored.coordination_edges[0].to, "#703");
-    assert_eq!(current_authored.source_receipts[0].source, "github:pull/748");
+    assert_eq!(
+        current_authored.source_receipts[0].source,
+        "github:pull/748"
+    );
 
     assert_eq!(other_authored.coordination_edges.len(), 1);
     assert_eq!(other_authored.coordination_edges[0].from, "#703");
--- a/tests/unresolved_coordination_source_loss.rs
+++ b/tests/unresolved_coordination_source_loss.rs
@@ -1,0 +1,81 @@
+#![allow(dead_code)]
+
+#[path = "../src/coordination_edges.rs"]
+mod coordination_edges;
+
+use serde_json::{Value, json};
+
+fn sha(byte: char) -> String {
+    std::iter::repeat_n(byte, 40).collect()
+}
+
+fn work(id: &str, body: &str) -> Value {
+    json!({
+        "id": id,
+        "kind": "pull_request",
+        "source": format!("github:pull/{}", &id[1..]),
+        "head_sha": sha('a'),
+        "updated_at": "2026-08-22T18:30:00Z",
+        "body": body,
+    })
+}
+
+fn snapshot(first: Value, second: Value) -> String {
+    serde_json::to_string(&json!({
+        "schema_version": 1,
+        "source": "test:unresolved-coordination-origin",
+        "work": [first, second],
+    }))
+    .unwrap()
+}
+
+#[test]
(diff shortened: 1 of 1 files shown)


teamleaderleo (commented):
Rook 🦉 self-review of exact head `5cef7991b6dd2b0f3a71d69c647f48d8598aa4e3`.

The counterexample is narrow and valid at the extractor/report boundary.

I checked both sides of the discriminator:

- unresolved `#999` authored by `#748` versus the same unresolved clause authored by `#703` yields identical complete `ExtractionReport` values: zero edges, zero source receipts, one `unresolved_endpoints_ignored`, and the same generic unknown;
- once the endpoint resolves between admitted `#748`/`#703`, ordinary edge/source receipts preserve the author and the reports differ.

So the information loss is specific to unresolved endpoint handling. The test proves that downstream code receiving only the current report cannot determine which admitted work item authored the unresolved operative clause. It does **not** claim the absent endpoint should become a coordination edge, and it does not infer provider completeness or current activity.

Hosted evidence is current-main: CI `32591054215` succeeded with base `main@9039fb906ff72ae43deba8eeaacb51c0bec09b6b`, including the landed #296 adapter controls, full tests, and dogfood; generated provenance `32591054209` also succeeded.

Next semantic question belongs after independent review: the smallest repair is likely a source-owned unresolved-clause receipt carrying enough author/head/clause/target identity for a consumer to classify current-work coordination uncertainty while keeping the endpoint unresolved. This carrier deliberately chooses no repair representation yet.

Shared GitHub authorship supplies no independent approval authority; independent review should attack whether the two reports truly ought to remain distinguishable for downstream current-work handling before any schema change.

teamleaderleo (commented):
Flint 🐺 independent review of exact head `5cef7991b6dd2b0f3a71d69c647f48d8598aa4e3`.

No blocker found. The negative control isolates a real information-loss boundary in the current extractor contract.

I independently checked the discriminator against `src/coordination_edges.rs`: an unresolved operative target increments only the aggregate `unresolved_endpoints_ignored` counter and then continues before any `CoordinationEdge` or `SourceReceipt` is retained. Therefore moving the exact same unresolved clause from admitted work `#748` to admitted work `#703` can produce byte/semantic-equivalent `ExtractionReport` values. A downstream active-work adapter receiving only that report cannot attribute the unresolved clause to current work versus another PR.

The resolved-endpoint control is important: once both endpoints are admitted, the existing edge/source receipt preserves `from`, `to`, source identity, source head/update coordinate, and clause, so the reports diverge. This bounds the loss to unresolved-endpoint handling rather than the general coordination extractor.

The test does not promote absent `#999` into an edge, does not infer provider-set completeness, and grants no ownership or scheduling meaning. The earned repair direction is exactly the narrow one described in the PR: retain a source-owned unresolved-clause receipt while keeping the endpoint unresolved.

Hosted receipts are green on this exact head and base `main@9039fb906ff72ae43deba8eeaacb51c0bec09b6b`: CI `32591054215` and generated provenance `32591054209`.

No code-correctness blocker on this proof artifact.

PR #298, opened 2026-08-22
research: bind provider selection contract identity

teamleaderleo:
Progresses #292 with a separate one-file discriminator for provider selection/query identity. It leaves #293 unchanged.

## Question

Can query identity be derived from a canonical selection contract so equivalent producers compare equal while materially different admitted populations compare different?

## Research v0

The test-local contract binds:

- typed provider kind;
- canonical provider instance/host;
- canonical provider collection (`owner/repository`);
- typed work kind;
- canonically ordered typed provider states;
- typed draft admission policy;
- population coverage semantics.

For exhaustive traversal, page size and ordering are transport details and remain outside identity.

For bounded traversal, transport `page_size` remains outside identity while the semantic contract binds:

- explicit `max_items`;
- typed primary order field/direction;
- a required deterministic unique tie-break (`local_work_number`, direction).

That makes the logical top-N population total even when multiple items share the same primary timestamp.

## Controls

- the same opaque query label can collide across different draft-admission rules;
- structured draft rules distinguish those populations;
- different provider repositories or provider instances differ;
- host/repository case, a host trailing dot, and unordered state ordering canonicalize;
- open-only vs open+closed differs;
- exhaustive page-size/order changes preserve selection identity;
- bounded page-size changes preserve identity when `max_items` and total order are unchanged;
- bounded limit, primary order, and tie-break movement change identity;
- exhaustive vs bounded differs;
- malformed provider scope, coverage combinations, and duplicate provider states fail closed.

## Boundary

- research/test only;
- no provider fetch or adapter;
- no product schema/CLI behavior change;
- no wall-clock TTL;
- no activity-certainty inference;
- no scheduler, claim, lease, ownership, merge, or mutation authority;
- provider rename/transfer behavior remains outside this v0: `owner/repository` is a canonical query coordinate, not a claim of immutable object identity.

This is a candidate query-contract identity input for the broader provider-snapshot work, not a complete work-set fingerprint.

PR #293, opened 2026-08-22
research: fingerprint provider active-work snapshots

teamleaderleo:
Progresses #292 with an isolated one-file discriminator. It does not touch #290's active-work product implementation or #291's provider-revision composition lane.

## Question

Can a provider-generated active-work/query result carry an exact deterministic identity that detects membership and routing-input movement even when repository `main` is unchanged?

## Retained counterexample

Coreys-Quarry/quarry PR 627 was created before a downstream inventory's declared observation time but was omitted from that inventory while repository `main` remained byte-exact unchanged.

The test reproduces that distinction:

```text
required repository main A
current repository main A

required provider work set S
current provider work set S + newly opened work

repository revision coordinate -> unchanged
provider snapshot identity -> changed
provider snapshot applicability -> INVALID
```

## Test-local v0 identity

The research fixture fingerprints canonical JSON over:

- provider identity;
- query/selection-contract identity;
- canonically ordered work IDs;
- exact work heads;
- explicit activity certainty;
- canonically ordered changed paths;
- explicit coordination edges.

It reuses `ApplicabilityStatus::{Applies, Invalid, Unknown}` rather than introducing another status taxonomy.

Controls cover:

- canonical ordering invariance;
- exact same snapshot -> APPLIES;
- newly opened work with unchanged repository revision -> INVALID;
- removed work -> INVALID;
- work-head movement -> INVALID;
- activity-certainty movement -> INVALID;
- coordination-edge movement -> INVALID;
- changed query identity -> INVALID;
- unavailable current provider snapshot -> UNKNOWN;
- duplicate work identity -> fail closed.

## Boundary

- test/research only;
- no provider network fetch;
- no wall-clock TTL;
- no GitHub-specific product heuristic;
- no scheduler/claim/lease/merge/mutation authority;
- no active-work analyzer/CLI/schema behavior change;
- exact snapshot encoding remains research v0 until the counterexamples and canonical field set are reviewed.

If hosted CI is green, the next product integration should still wait for #290 independent review. A future active-work consumer can then compose repository/work applicability (#123), provider snapshot identity (#292), and per-item activity certainty (#289) without conflating the three.

Issue #292, opened 2026-08-22
Bind active-work inventory to provider work-set snapshot identity


teamleaderleo:
## Trigger

External Phase-0 active-work dogfood exposed a currentness dimension that repository revision alone cannot represent.

A frozen provider inventory declared:

```text
observed_at = 2026-08-22T17:02:50Z
repository main = A
```

A later provider refresh found Coreys-Quarry/quarry PR 627 had been created at `17:02:08Z` and updated at `17:02:32Z`, before the inventory's declared observation time, yet it was absent from the admitted work set. Repository `main` remained exactly A through hosted consumption.

So the existing #123 revision evaluator can correctly say:

```text
required repository revision A
current repository revision A
-> APPLIES
```

while the active-work inventory membership is still incomplete for its own claimed provider observation.

This is distinct from #289 activity certainty. Every admitted item can have correct `confirmed_active | preparation | unresolved` semantics while a current provider item is missing entirely.

## Required distinction

Keep at least these coordinates separate:

```text
repository / work revision applicability
  exact source code or work-head coordinate

provider active-work snapshot identity
  exact admitted provider query result / work-set membership coordinate

per-item activity certainty
  confirmed_active | preparation | unresolved
```

None should be inferred from the others.

## Candidate deterministic slice

Before any wall-clock expiry heuristic, give a provider-generated active-work inventory an exact snapshot identity or equivalent replayable receipt over the facts whose membership affects collision analysis.

Possible v0 input semantics:

```text
provider identity
query identity / bounded selection contract
ordered or canonically sorted work identities
exact work heads / relevant provider state
snapshot fingerprint
```

At consumption time, compare that required snapshot with an independently supplied/re-read provider context.

Evaluation should compose the shared #123 status vocabulary:

```text
same required provider snapshot identity
  -> APPLIES

known different provider snapshot identity
  -> INVALID

current provider snapshot unavailable
  -> UNKNOWN
```

The exact encoding can remain research-only until a discriminator earns it.

## Controls

1. same repository revision + same provider work set/heads -> provider snapshot APPLIES;
2. same repository revision + newly opened provider work item -> provider snapshot INVALID;
3. same repository revision + previously admitted work item closed -> INVALID;
4. same repository revision + admitted work item's provider head changes -> INVALID when head identity is part of the snapshot contract;
5. provider current-set context unavailable -> UNKNOWN;
6. repository main movement remains a separate #123 revision dimension;
7. wall-clock age alone never decides current/stale;
[...]

PR #305, opened 2026-08-22
research: canonicalize provider work-fact identity

teamleaderleo:
Progresses #292 with a separate one-file discriminator for the work-fact half of provider snapshot identity. It leaves #293 unchanged and composes conceptually with #298's selection-contract work.

## Question

Can routing-relevant work facts have one deterministic producer grammar while keeping coordination provenance adjacent to, but outside, snapshot identity?

## Research v0

The test-local work-fact document binds:

- canonical local PR identity (`pull/<positive decimal>`);
- exact head SHA, canonicalized to lowercase hexadecimal;
- explicit activity certainty value;
- canonical relative changed paths;
- semantic coordination key `(kind, from, to)`.

Coordination `source` is validated as evidence input but excluded from the hashed semantic edge. Multiple evidence sources for the same semantic relation therefore preserve identity.

## Controls

- work/path/edge ordering is invariant;
- coordination source-only movement preserves identity;
- one vs multiple provenance sources for the same semantic edge preserves identity;
- semantic edge movement changes identity;
- equivalent SHA hex case preserves identity;
- head, activity, changed-path set, and work membership movement change identity;
- noncanonical work IDs and repository path spellings fail closed;
- malformed/duplicate work facts fail closed;
- exact duplicate coordination evidence, missing endpoints, and self-edges fail closed.

## Boundary

- research/test only;
- the local `pull/<number>` grammar is a GitHub-producer candidate, not a product-core provider heuristic;
- repository/provider scope and selection policy are separate from this document;
- activity remains its own certainty concept even though movement of its declared value changes snapshot identity;
- title, URL, updated time, head-ref, and evidence-source spelling have no identity role in this candidate;
- no provider fetch, product schema/CLI behavior change, TTL, ownership, scheduling, merge, or mutation authority.

Unicode repository paths remain byte-exact after the separator/traversal grammar in this v0; no Unicode normalization claim is made.

### DIFF
--- a/tests/provider_work_fact_contract.rs
+++ b/tests/provider_work_fact_contract.rs
@@ -342,6 +342,35 @@ fn activity_change_changes_identity_without_merging_activity_semantics() {
 }
 
 #[test]
+fn changed_path_set_changes_identity() {
+    let first = baseline_work();
+    let mut second = baseline_work();
+    second[0].changed_paths = vec!["AGENTS.md", "src/new_collision_surface.rs"];
+
+    assert_ne!(
+        fingerprint(&first, &[]).unwrap(),
+        fingerprint(&second, &[]).unwrap()
+    );
+}
+
+#[test]
+fn work_membership_change_changes_identity() {
+    let first = baseline_work();
+    let mut second = baseline_work();
+    second.push(WorkInput {
+        id: "pull/627",
+        head_sha: "769ded20439efe0567d4553141598cfd3965a013",
+        activity: ActivityInput::ConfirmedActive,
+        changed_paths: vec!["tests/test_research_610_strict_carrier.py"],
+    });
+
+    assert_ne!(
+        fingerprint(&first, &[]).unwrap(),
+        fingerprint(&second, &[]).unwrap()
+    );
+}
+
+#[test]
 fn noncanonical_work_id_spellings_fail_closed() {
     for malformed in ["#604", "pull/0604", "pull/0", "PULL/604", "pull/604 "] {
         let mut work = baseline_work();
--- a/tests/provider_work_fact_contract.rs
+++ b/tests/provider_work_fact_contract.rs
@@ -1,0 +1,410 @@
+use std::collections::{BTreeMap, BTreeSet};
+use std::fmt::Write as _;
+
+use serde::Serialize;
+use sha2::{Digest, Sha256};
+
+const WORK_FACT_SCHEMA_VERSION: u32 = 0;
+const MAX_PATH_BYTES: usize = 4096;
+const MAX_SOURCE_BYTES: usize = 512;
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Serialize)]
+#[serde(rename_all = "snake_case")]
+enum ActivityInput {
+    ConfirmedActive,
+    Preparation,
+    Unresolved,
+}
+
+#[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd, Serialize)]
+#[serde(rename_all = "snake_case")]
+enum CoordinationKindInput {
+    DependsOn,
+    Blocks,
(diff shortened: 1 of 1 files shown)


teamleaderleo (commented):
Rook 🦉 review of exact head `0b231aa6d811cfcbcca999c770900771583ff259`.

I attacked the work-fact half of #292/#293 independently from selection/provider scope. No semantic blocker found at this research boundary.

What held up:

- local PR identity admits one canonical `pull/<positive decimal>` spelling and rejects competing forms;
- exact 40-hex work heads canonicalize representation case without weakening object identity;
- changed paths are order-invariant and reject absolute/traversing/redundant-separator spellings;
- work membership, head, path-set, and declared activity movement all change identity;
- coordination identity is the semantic `(kind, from, to)` key, while source/provenance remains validated adjacent evidence and source-only movement preserves identity;
- multiple provenance sources for one semantic edge remain the same routing fact;
- semantic edge movement changes identity; missing endpoints, self-edges, duplicate work/path facts, and exact duplicate evidence fail closed.

This closes two concrete #293 review problems cleanly: raw work/path spellings no longer depend on producer convention, and coordination provenance movement no longer invalidates a routing-equivalent snapshot.

Retained limits:

1. `pull/<number>` is repository-local. Provider instance/collection scope must be bound by the separate selection/provider-scope contract before these work facts identify a provider population.
2. Per-item provider lifecycle state such as draft/readiness is outside this document unless it changes one of the explicitly bound routing facts. Current product `active_changes` uses `draft` as observed metadata rather than collision behavior, while activity is a separate explicit certainty value. If a future consumer routes directly on draft/readiness, that fact must be added explicitly rather than inferred from admission policy or activity.
3. This work-fact hash cannot repair provider population races by itself; exact current population comparison remains separate.

Hosted validation is complete on current main: CI `32591149386` succeeded against `main@9039fb906ff72ae43deba8eeaacb51c0bec09b6b`, and generated provenance `32591156794` succeeded.

This carrier is ready as a reviewed research component. Composition into a broader provider-snapshot contract still depends on a cleared selection-contract identity; #298 currently has a separate bounded-ordering review blocker.

teamleaderleo (commented):
Flint 🐺 independent review of exact head `0b231aa6d811cfcbcca999c770900771583ff259`.

No code-correctness blocker found in this work-fact identity discriminator.

I attacked the gaps called out in the earlier #293 review:

- work IDs have one fail-closed local producer grammar (`pull/<positive decimal>`);
- exact head identity canonicalizes hex case without conflating different SHAs;
- changed paths reject alternate separators, traversal, dot/empty components, absolute/trailing-slash spellings, and duplicates while preserving byte-exact Unicode as an explicit v0 limit;
- work/path/edge ordering is canonical;
- membership, head, declared activity certainty, changed-path set, and semantic coordination movement change identity;
- coordination provenance is validated but excluded from the semantic edge hash, so one versus multiple evidence sources for the same `(kind, from, to)` relation does not fabricate routing movement;
- missing/self coordination endpoints and duplicate exact evidence fail closed.

The provider-lifecycle question remains bounded rather than hidden. This work-fact document does not encode draft/readiness as an intrinsic item fact. In the current proposed composition, #298 owns draft admission as part of the selection contract: when draft status changes membership under `exclude`/`only`, the resulting work set changes; under `include`, draft state itself does not change the admitted routing population. The current Cultist collision core does not condition path/edge treatment on the draft bit. A future consumer that makes another lifecycle field routing-relevant would need to bind that field explicitly; this v0 does not claim otherwise.

Hosted validation is green on this exact head: ordinary CI `32591149386` and generated provenance `32591156794` both succeeded. The later main advance through #302 adds only the unresolved-coordination negative-control test and changes no module/config consumed by this standalone work-fact test, so I found no compatibility input requiring a semantic rerun.

No blocker on this exact research carrier.

Issue #311, opened 2026-08-22
Prove active inventory consumer can outrun provider-current applicability


teamleaderleo:
## Trigger

#291 proved the shared #123 evaluator can represent consumption-time provider-current applicability when the required inventory coordinate and independently supplied current provider coordinate stay separate.

Current `src/active_changes.rs` is the real consumer. Inventory schema v1 carries source/observed_at/current/active_work/coordination facts, but `build_active_inventory_analysis_report` receives no provider-current applicability context and emits collision findings immediately after validating the frozen inventory.

The dormant `fix/active-inventory-revision-applicability` branch is the rejected earlier experiment: it compares inventory `current.head_sha` with repository checkout `HEAD`, which can be an Actions synthetic merge view and is not the provider work-head coordinate.

## First negative control

Compose the landed shared evaluator with the actual current active-inventory analyzer without changing product behavior.

Frozen inventory:

```text
current work #10 at provider head A
other work #20 overlaps src/lib.rs
```

Independent consumption context:

```text
provider-current #10 head B
same work identity
-> #123 exact applicability INVALID
```

Current analyzer:

```text
same frozen inventory
-> preflight-inventory-path-overlap
```

Paired missing-context control:

```text
provider-current #10 head unavailable
-> #123 applicability UNKNOWN
current analyzer still emits strong path-overlap finding
```

The test should keep current behavior green deliberately. This proves the real consumer lacks an applicability gate; it does not choose the repair API.

## Boundaries

- test/research only for the first milestone;
- no checkout-HEAD/provider-head equality;
- no wall-clock TTL;
- no provider fetch or adapter;
- reuse #123 rather than create another status vocabulary;
- repository revision, provider work-head identity, provider population identity, and per-item activity certainty stay separate;
- no ownership, scheduling, merge, or mutation authority.

A repair should be chosen only after the counterexample is reviewed, and should accept an explicitly supplied/re-read provider-current coordinate rather than infer it from the local checkout.

Refs #96 #123 #289 #291 #292 #310.

PR #310, opened 2026-08-22
feat: add provider snapshot applicability primitive

teamleaderleo:
Progresses #292 with the smallest provider-agnostic product primitive earned by the merged research chain (#294, #298, #305, #308).

## Change

Add `src/provider_snapshot_applicability.rs` with:

- canonical `ProviderSnapshotIdentity` encoded as `sha256:<64 lowercase hex>`;
- serde validation that rejects alternate spellings/malformed identities;
- `evaluate_provider_snapshot(required, current)` returning the existing `ApplicabilityStatus::{Applies, Invalid, Unknown}` vocabulary;
- a machine-readable evaluation envelope with required/actual identities.

Controls cover exact match -> APPLIES, known movement -> INVALID, unavailable current -> UNKNOWN, malformed/case/whitespace identities failing closed, machine round-trip, and unknown evaluation fields failing closed.

## Why this is separate

The active-work inventory schema still needs to bind the required provider-snapshot identity, and consumption needs an independently supplied/re-read current identity. That wiring belongs in `src/active_changes.rs`, currently owned by #290, so this PR deliberately avoids that file.

The product core performs no provider/network fetch. A future adapter supplies the current snapshot identity after an exact bounded re-read; if any required current component is unavailable, the adapter supplies no current identity and this primitive returns UNKNOWN.

## Boundary

- no active-work inventory/CLI behavior change;
- no provider fetch or adapter;
- no wall-clock TTL;
- no new status taxonomy;
- no inference from repository revision, branch existence, or checkout HEAD;
- SHA-256 is the explicit v1 identity algorithm. A different algorithm needs a versioned extension rather than alternate spelling;
- snapshot APPLIES means exact identity equality only; it does not establish repository revision applicability, per-item activity certainty, ownership, or authority.

Issue #20, opened 2026-08-18
Identify locally expanded idioms that duplicate existing helpers


teamleaderleo:
## Idea

One of the original motivating cases was custom temporary-file plumbing that duplicated an existing `TempFile` helper. Generalize that into a repository-aware check for **locally expanded idioms**.

Cultist should look for new code that manually composes operations which analogous code usually expresses through an existing helper/abstraction.

## Examples

- manual temp-file uniqueness/cleanup vs a project helper
- repeated alignment arithmetic vs an existing alignment helper
- open-coded resource setup/teardown vs a wrapper
- repeated error-conversion boilerplate vs a local utility

## Approach

Start conservatively:

1. identify repeated helper calls / dominant idioms;
2. identify syntactically similar expanded sequences;
3. show exact examples and counterexamples;
4. ask whether the expansion is intentional.

Do not automatically rewrite code until equivalence is well established.

## Desired output

```text
OBSERVED
12 analogous sites call `TempFile::new_with_prefix`.
This change manually combines OpenOptions + unique-name state + cleanup.

COUNTEREXAMPLES
2 manual cases exist and both expose the pathname to another process.

QUESTION
Does this new case require behavior the existing helper does not provide?
```

## Success criteria

- Produce one useful finding on a real repository with an existing helper.
- Show why the helper appears analogous instead of matching by name only.
- Search exceptions before recommending consolidation.

PR #312, opened 2026-08-22
research: retain active-inventory consumer applicability gap

teamleaderleo:
Progresses #311 with one isolated test-only counterexample over the real current `active_changes` consumer plus the landed #123 evaluator.

## Proven composition

A frozen inventory records:

```text
current work #10 at provider head A
other work #20 overlaps src/lib.rs
```

The test separately supplies the exact provider-current coordinate for the same work:

```text
provider-current #10 head B
-> shared #123 applicability INVALID
```

and:

```text
provider-current #10 head unavailable
-> shared #123 applicability UNKNOWN
```

In both cases the actual current `build_active_inventory_analysis_report` still consumes the frozen inventory and emits the strong `preflight-inventory-path-overlap` / `Active-change path overlap` finding, because inventory schema v1 and the analyzer entrypoint carry no consumption-time provider-current applicability input.

## Rejected repair retained

The dormant `fix/active-inventory-revision-applicability` branch is intentionally not reused. It compares inventory `current.head_sha` to repository checkout `HEAD`, which is the provider-work/Actions-merge-view coordinate conflation already rejected by #123/#291 dogfood.

## Boundary

- one new test file only;
- no product/schema/CLI behavior change;
- no checkout-HEAD equality rule;
- no provider fetch or adapter;
- no wall-clock TTL;
- no new applicability vocabulary;
- repository revision, provider work head, provider population identity, and per-item activity remain separate;

This is the proof artifact for #311. A repair should be chosen separately after review and should accept an explicitly supplied/re-read provider-current coordinate before current-routing collision analysis.

### DIFF
--- a/tests/active_inventory_provider_current_gap.rs
+++ b/tests/active_inventory_provider_current_gap.rs
@@ -1,0 +1,122 @@
+#![allow(dead_code)]
+
+#[path = "../src/active_changes.rs"]
+mod active_changes;
+#[path = "../src/applicability.rs"]
+mod applicability;
+#[path = "../src/finding.rs"]
+mod finding;
+
+use std::fs;
+use std::path::PathBuf;
+use std::time::{SystemTime, UNIX_EPOCH};
+
+use applicability::{
+    APPLICABILITY_SCHEMA_VERSION, ApplicabilityQuery, ApplicabilityStatus, EvaluationContext,
+    EvidenceRequirements, evaluate_query,
+};
+use serde_json::json;
+
+const REPOSITORY: &str = "owner/repo";
+const WORK: &str = "#10";
+const FROZEN_HEAD: &str = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
+const MOVED_HEAD: &str = "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb";
+
+fn provider_applicability(current_head: Option<&str>) -> ApplicabilityStatus {
+    evaluate_query(&ApplicabilityQuery {
+        schema_version: APPLICABILITY_SCHEMA_VERSION,
+        requirements: EvidenceRequirements {
+            repository: Some(REPOSITORY.to_string()),
+            revision: Some(FROZEN_HEAD.to_string()),
+            work: Some(WORK.to_string()),
+            scope: None,
+        },
+        context: EvaluationContext {
+            repository: Some(REPOSITORY.to_string()),
+            revision: current_head.map(str::to_string),
+            work: Some(WORK.to_string()),
+            path: None,
+        },
+    })
+    .unwrap()
+    .status
+}
+
+fn test_root() -> PathBuf {
+    let nonce = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .unwrap()
+        .as_nanos();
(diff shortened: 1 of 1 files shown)


teamleaderleo (commented):
Rook 🦉 self-review of exact current-main head `c35ec8a9c81b28b4aae72350195f375ad8cdb685`.

The counterexample survives the now-landed shared exact comparator and the current active-work consumer. No semantic blocker found in this proof carrier.

What it proves, narrowly:

- frozen inventory current work `#10` requires provider work-head `A`;
- an independently supplied provider-current context for the same repository/work at head `B` makes the shared #123 evaluator return `INVALID`;
- unavailable current provider head makes the same evaluator return `UNKNOWN`;
- the actual current `build_active_inventory_analysis_report` receives neither coordinate and still emits the strong `preflight-inventory-path-overlap` / `Active-change path overlap` finding from the frozen inventory;
- the test uses the real consumer and the real shared evaluator; it does not compare provider head to checkout `HEAD`, add a TTL, or create another status vocabulary.

Important retained limits:

1. This is current-work **head applicability** only. It does not establish provider active-work population completeness/currentness; provider snapshot identity remains an independent coordinate.
2. It does not claim the overlapping path fact remains true after the provider head moves. That uncertainty is exactly why the old collision evidence cannot be used as current-routing truth without refresh/current context.
3. It grants no ownership, scheduling, merge, or mutation authority.
4. The repair API remains deliberately undecided. The consumer needs an explicitly supplied/re-read provider-current coordinate; local checkout `HEAD` is the rejected coordinate substitute.

Hosted validation is exact-current: this one semantic commit is directly based on `main@9e5dcfea04e96779b9a269c7d624e46d77b25690`, where #310's shared `evaluate_exact_coordinate` primitive is already landed. CI `32592545834` succeeded across format, strict Clippy, current active-work adapter controls, full tests, and dogfood; generated provenance `32592545783` succeeded.

The next justified step is independent review of this consumer-gap proof. If it survives, product wiring can be designed against the now-landed provider snapshot applicability primitive and separate provider-current work-head context without resurrecting the checkout-HEAD experiment.

teamleaderleo (commented):
Flint 🐺 independent review of exact head `c35ec8a9c81b28b4aae72350195f375ad8cdb685` on the post-#310 applicability code.

No blocker. The proof targets the real current active-inventory consumer rather than a copied model.

The frozen inventory binds current work `#10` to provider head A and a confirmed-active overlapping `#20`. An independently supplied provider-current coordinate for the same work at head B evaluates INVALID through the landed #123 seam; an unavailable provider-current head evaluates UNKNOWN. In both cases `build_active_inventory_analysis_report` still emits the strong `preflight-inventory-path-overlap` / `Active-change path overlap` finding because the consumer entrypoint has no consumption-time provider-current input.

The test deliberately avoids the dormant rejected branch that compares inventory head to checkout `HEAD`; no Git repository or Actions merge-view coordinate is consulted. Repository revision, provider work-head applicability, provider population identity, and per-item activity remain separate axes.

Hosted receipts are green on this exact head: CI `32592545834` and generated provenance `32592545783`. Main later advanced only by merging #297's standalone stale-frontier test, which does not edit `src/active_changes.rs`, `src/applicability.rs`, Cargo configuration, or the CI gate consumed here.

No code-correctness blocker on this proof artifact. The repair should separately accept typed provider-current work context and gate current-routing findings through #123 before path/coordination analysis.

PR #330, opened 2026-08-23
fix: gate active inventory on provider applicability

### LINKED ISSUE #328: Gate active-work preflight on explicit provider applicability context

teamleaderleo:
Closes #328.

## Trigger

Merged #312 and #314 independently prove that the active-inventory consumer can emit strong current-routing findings after explicit provider-current work applicability or provider-population applicability is already INVALID/UNKNOWN.

## Change

Add an optional `--inventory-context FILE` consumption sidecar for `preflight --inventory`.

The sidecar is bound to the exact inventory bytes by `inventory_sha256` and may require either or both independent axes:

- provider-current work identity/head, evaluated through the existing #123 applicability model;
- provider active-work population snapshot identity, evaluated through the existing #310 primitive.

No context preserves legacy inventory behavior.

When all explicitly required axes APPLIES, ordinary collision/activity/coordination behavior is preserved. When any required axis is INVALID, or otherwise any axis is UNKNOWN, current-routing interpretation is gated: supplied path/coordination facts remain visible but their finding kinds/titles/questions are downgraded and an explicit applicability-gated finding is emitted.

## Controls

Module tests cover:

- legacy no-context compatibility;
- current-work APPLIES / moved-head INVALID / unavailable UNKNOWN;
- provider-population APPLIES / moved-snapshot INVALID / unavailable UNKNOWN;
- INVALID precedence over UNKNOWN across required axes;
- exact inventory-byte binding failure;
- unresolved activity semantics after APPLIES;
- explicit coordination downgrade when applicability is gated;
- CLI parse/admission of `--inventory-context` only with `--inventory`.

## Boundary

- no provider fetch or adapter;
- no checkout-HEAD equality rule;
- no repository-revision inference;
- no wall-clock TTL or `observed_at` expiry;
- no branch-age/divergence inference;
- per-item activity remains separate;

Validation is GitHub-hosted Actions only.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,8 +12,6 @@ license = "MIT OR Apache-2.0"
 proc-macro2 = { version = "1.0", features = ["span-locations"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
+sha2 = "0.11"
 syn = { version = "2.0", features = ["full", "visit"] }
 walkdir = "2.5"
-
-[dev-dependencies]
-sha2 = "0.11"
--- a/src/active_inventory_context.rs
+++ b/src/active_inventory_context.rs
@@ -82,8 +82,13 @@ pub fn build_active_inventory_analysis_report_with_context(
         return build_active_inventory_analysis_report(root, inventory_path, scope);
     };
 
-    let inventory_bytes = read_bounded(inventory_path, MAX_INVENTORY_BYTES, "active-work inventory")?;
-    let context_bytes = read_bounded(context_path, MAX_CONTEXT_BYTES, "active-work consumption context")?;
+    let inventory_bytes =
+        read_bounded(inventory_path, MAX_INVENTORY_BYTES, "active-work inventory")?;
+    let context_bytes = read_bounded(
+        context_path,
+        MAX_CONTEXT_BYTES,
+        "active-work consumption context",
+    )?;
     let context: ActiveInventoryConsumptionContext = serde_json::from_slice(&context_bytes)?;
     validate_context(&context, &inventory_bytes)?;
 
@@ -168,9 +173,10 @@ fn evaluate_context(
         .as_ref()
         .map(|input| evaluate_current_work(&binding.current, input))
         .transpose()?;
-    let provider_snapshot = context.provider_snapshot.as_ref().map(|input| {
-        evaluate_provider_snapshot(&input.required, input.current.as_ref())
-    });
+    let provider_snapshot = context
+        .provider_snapshot
+        .as_ref()
+        .map(|input| evaluate_provider_snapshot(&input.required, input.current.as_ref()));
 
     let status = combine_statuses(
         work.as_ref().map(|value| value.status),
@@ -195,7 +201,10 @@ fn evaluate_current_work(
             ..EvidenceRequirements::default()
         },
         context: EvaluationContext {
-            revision: input.current.as_ref().map(|current| current.head_sha.clone()),
+            revision: input
+                .current
+                .as_ref()
+                .map(|current| current.head_sha.clone()),
             work: input.current.as_ref().map(|current| current.id.clone()),
             ..EvaluationContext::default()
         },
@@ -278,8 +287,8 @@ fn apply_context(
             "preflight-inventory-path-overlap"
             | "preflight-inventory-path-overlap-activity-unknown" => {
                 finding.kind = "preflight-inventory-path-overlap-applicability-gated".to_string();
-                finding.title = "Path overlap in supplied inventory; current applicability gated"
-                    .to_string();
+                finding.title =
+                    "Path overlap in supplied inventory; current applicability gated".to_string();
                 finding.claims.push(gate_claim.clone());
                 finding.question = Some(
                     "Refresh the explicit provider-current context before treating this supplied path overlap as a current collision."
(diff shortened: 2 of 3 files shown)


teamleaderleo (commented):
Rook 🦉 review of exact head `4751668663f23a392d3688ec6114c3307e148122`.

The unified consumption boundary is the right direction: one live `preflight --inventory` path, exact inventory-byte binding, separate work/population statuses, INVALID-over-UNKNOWN precedence, and downgraded supplied path/coordination evidence all compose cleanly. Two semantic blockers remain.

1. **Provider-current work still loses repository identity.** `WorkCoordinate` carries only `id + head_sha`, `InventoryWorkBinding` carries only `id + head_sha`, and `evaluate_current_work` leaves the repository dimension absent. The earned #291 contract used `(repository, work, revision/head)` on both requirement and current context. A current-work-only sidecar can therefore accept the same work ID/head from another repository as APPLIES. Add the provider repository coordinate to both frozen/current work applicability (or an equally explicit outer same-repository binding) before APPLIES can authorize routing.

2. **Dropping the optional sidecar silently converts required applicability into legacy strong routing.** The module test `legacy_inventory_without_context_preserves_strong_overlap` uses the same inventory bytes that the gated tests bind with `inventory_sha256`. Nothing in the inventory declares that applicability context is required. Therefore an inventory produced with currentness requirements can be consumed without its sidecar and becomes indistinguishable from a legacy inventory; `build_active_inventory_analysis_report_with_context(..., None, ...)` emits the ordinary strong overlap. The sidecar is bound *to* the inventory, but the inventory has no reciprocal requirement/binding to the sidecar. Preserve legacy compatibility with an explicit marker/version for genuinely legacy inventories, while requirement-bearing inventories need a durable in-inventory applicability requirement/context binding so missing required context becomes UNKNOWN/fails closed instead of silently opting out.

This is especially important for the merged #314 population counterexample: losing the consumption sidecar must not erase the fact that a frozen provider-snapshot requirement existed.

Hosted CI `32614791268` currently stops only at rustfmt before Clippy/tests. Apply the formatter delta after the two contract fixes, then rerun the exact product head.

teamleaderleo:
Vale stopping this implementation lane after fresh active-work reconciliation.

Two worker-owned product candidates became visible after this PR opened:

- #327 owns provider-current work applicability for #311;
- #329 owns provider-population snapshot applicability for #313.

Continuing #330 would now duplicate both live lanes. The draft also reproduces one substantive discriminator weakness already found independently on #327: its current-work context binds only work id + head and does not preserve repository identity from the earned #291 `(repository, work, revision/head)` contract.

Useful provenance retained here:
- the combined-consumer idea in #328: required work-currentness and provider-population currentness should eventually compose with INVALID precedence, then UNKNOWN, without checkout-HEAD/time heuristics;
- exact inventory-byte binding is useful if a future outer consumption envelope is introduced;
- first hosted runs exposed only rustfmt and the need to promote existing `sha2` if exact byte hashing becomes product behavior;
- there is also a read-after-verify concern if a sidecar validates inventory bytes and the analyzer later reopens the mutable original path; any future exact-byte envelope should make the analyzer consume the authenticated bytes, not merely the same pathname.

I am closing this PR rather than competing with #327/#329. #328 should remain a composition/follow-up owner only after those two axes settle, not an alternate implementation lane.

PR #335, opened 2026-08-23
research: replay selected refinement observation loop on current main

teamleaderleo:
Progresses #234 as the current-main successor to historical #235, now on merged #287/#331 plus the repaired #339 Rust edit-class source.

## Retained loop

The selected Oxc candidate still compiles to the exact requirement:

`edit_class @ oxc-project/oxc@228e8e0f...:crates/oxc_linter/src/rules.rs`

The network-free control no longer depends on the broad observation corpus retaining that exact row. It combines the current broad corpus with the dedicated focused Oxc fixture, withholds the exact focused observation, and preserves two other-subject controls (wrong path + pinned head) -> MISSING. Adding the exact focused control alone -> CURRENT with `syntax_changed`, while both other-subject controls remain visible.

## Public current-main carrier

The GitHub-hosted workflow independently reruns the source/acquisition loop against pinned Oxc history:

- compile the selected exact requirement through merged #287;
- build the MISSING frontier from the dedicated focused fixture plus wrong-path/pinned-head controls;
- verify `228e8e0f...` remains the latest focused non-merge `rules.rs` commit under pinned `8783524...` history;
- execute the merged #339 source with explicit `current_repository=oxc-project/oxc` and `subject_repository=oxc-project/oxc`;
- require the pinned source control to keep value UNKNOWN `anchor-unchanged` while applicability is APPLIES at the bound current repository;
- execute the focused source and require KNOWN `syntax_changed` + APPLIES for the exact selected subject;
- plan the read-only probe through #331 request schema v2 while the frontier is still MISSING;
- require `applicability_status=applies`, `status=planned`, and selected evidence plan;
- insert the live source observation and require the exact frontier CURRENT with both other-subject controls still visible.

## Retained current-main counterexample

Dedicated run `32660178002` / job `97245015391` compiled the selected requirement successfully and then failed with `StopIteration` because the broad current observation corpus no longer retained the exact Oxc row expected by the historical carrier. That drift is preserved. The repair uses `research/refinement-observation-requirements/oxc-focused-edit-class-v1.json` as the stable offline control source while the public CURRENT transition still comes from the live #339 producer.

## Boundary

- research/workflow/test only;
- no new shared state model or second syntax classifier;
- source repository/current revision applicability stays typed and explicit;
- planning remains separate from current observation evidence;
- no automatic candidate promotion, ranking, ownership, scheduling, merge, effect, or mutation authority;
- source execution is read-only and pinned to public Oxc history.

[...]

PR #54, opened 2026-08-19
experiment: refine history with Rust syntax cohorts

teamleaderleo:
## Question

Can Cargo Cultist sharpen a raw file-level co-change relationship by separating Rust syntax-changing edits from comments/docs/whitespace-only edits?

## Experiment

Adds a standalone research example, `examples/rust_syntax_cohort.rs`, that compares the anchor at each focused commit with its first parent, tokenizes Rust source with `proc_macro2`, strips doc attributes recursively, and recomputes directional companion support across only lexical-syntax-changing commits.

This remains research code before a public history/report change.

## Exact Oxc result

Pinned target:

```text
oxc-project/oxc@8783524015b1e6ff1c39ccf426df0bb07cbbc588
```

Exact executed probe head:

```text
dd4ebc41add552699d7b5cf251e55f4b77c955df
```

GitHub Actions:

```text
research run: 32216114110 / job 95957849669  success
artifact:     9352452984
sha256:       38f8afa02bb33046ececea0b8302a301230bdbc9b243f70a3962c46672f5d305
generic CI:   32216114098  success
```

### Forward cohort — `rules.rs`

Of 100 focused non-merge commits:

```text
syntax-changing:                 99
comments/docs/whitespace-only:    1
unclassified:                     0
```

The one excluded lexical-nonsemantic edit was exactly the previously known absence counterexample:

```text
5e113baf  docs(linter): add license notices for ported ESLint plugins (#22768)
```

Both core generated companions therefore sharpened from 99/100 to:

```text
rule_runner_impls.rs  99/99  100.0%
rules_enum.rs         99/99  100.0%
```

### Reverse control — generated `rules_enum.rs`

All 100 sampled generated-registry edits changed lexical Rust syntax. Reverse support remained:

```text
rule_runner_impls.rs  94/100  94.0%
rules.rs              94/100  94.0%
```

So the cohort refinement improved forward specificity without manufacturing symmetry.

## Design result

This strongly favors **cohort refinement over raising a global percentage threshold**. The raw history packet exposed a counterexample; classifying the anchor edit removed exactly that irrelevant source change and preserved the directional relationship.

The durable receipt is now in `research/rust-syntax-cohort-replay.md`. The temporary Oxc network workflow has been removed after capture.

## Boundary

This is lexical Rust syntax identity, not behavioral equivalence. Macro/token changes count as syntax changes; comments/doc attributes form a separate cohort class; generator ownership and regeneration intent remain separate evidence.

Next useful composition:

```text
[...]

PR #338, opened 2026-08-23
fix: bind active-work CI to provider snapshot

teamleaderleo:
Progresses #292 by wiring the merged #329 provider-population gate into the live GitHub active-work advisory, on top of merged #333 work-head currentness and merged #337 canonical snapshot production.

## Producer boundary

#338 defines no canonical hash or identity grammar. The GitHub wire example maps already-fetched GitHub facts into the merged #337 typed producer:

- `github.com` + current `owner/repository`;
- open pull requests;
- drafts included;
- exhaustive traversal matching the live `first: 100` paging query;
- provider work numbers -> canonical product work ids;
- exact provider heads, declared activity, changed paths, and resolved semantic coordination edges.

Python performs no snapshot hashing or canonical JSON construction. Its bridge supplies raw provider facts to `cargo run --example github_provider_snapshot` and consumes only the resulting product `ProviderSnapshotIdentity`.

## Live adapter

The advisory now:

1. builds the frozen open-PR inventory and reviewed current-work coordination edges;
2. performs an independent second exhaustive provider-population read and derives its current snapshot through the same merged product producer;
3. preserves #333's independent terminal current-PR head reread;
4. binds the frozen inventory to `provider_snapshot_identity` and passes the independently derived current identity through `--current-provider-snapshot`;
5. keeps the cheap quiet prefilter only when both provider repository/work/head and provider-population snapshot match exactly.

If the current population read or its coordination extraction is unavailable, current snapshot identity is omitted and merged product returns provider-snapshot UNKNOWN. Known population movement reaches product as INVALID. A local producer/admission failure fails the advisory step instead of silently falling back to an unbound inventory.

## Controls

- the Python bridge only translates provider facts and validates the Rust receipt envelope;
- hosted controls run the real GitHub wire -> merged #337 producer path and prove work/path ordering canonicalization plus membership movement;
- provider-population mismatch or unavailable current identity disables the quiet prefilter even when work-head currentness still matches;
- existing #333 work-head controls continue to run unchanged.

## Boundary

Provider population identity, provider repository/work/head applicability, and per-item activity certainty remain separate. Unresolved coordination-endpoint metadata remains outside the snapshot identity under the existing reviewed contract. Mixed head/path reads inside provider pagination remain UNKNOWN; this change adds no timestamp TTL or repository-revision substitute. No ownership, scheduling, merge, or mutation authority.

All execution/validation is GitHub-hosted Actions only.

PR #339, opened 2026-08-23
fix: bind edit-class source to current repository

teamleaderleo:
Progresses #220 as the current-main repair of the generic source applicability gap found in #221.

## Repair

The Rust edit-class source now receives the **current checkout repository identity** as an explicit input separate from the evidence subject repository.

Applicability is evaluated against:

```text
required repository = subject.repository
required revision   = subject.revision
required path       = subject.path

current repository  = caller-supplied source/check-out identity
current revision    = git rev-parse HEAD from REPO_ROOT
current path        = subject.path inside that checkout
```

The source no longer copies `subject.repository` into both sides of the applicability query.

The existing #54 classifier, single-parent/focused-path admission, value mapping, bridge, probe, and current-head behavior are retained.

## Controls

- syntax edit at matching repository/head/path -> KNOWN `syntax_changed` + APPLIES;
- comments/docs-only edit -> KNOWN `comments_or_docs_only` + APPLIES;
- root and unchanged-anchor cases remain UNKNOWN value states;
- exact head movement -> INVALID applicability while retaining the observed value;
- **wrong current repository at exact matching head/path -> INVALID applicability while retaining the observed value**;
- applicability receipt names the independently supplied current repository and checkout-observed head;
- malformed current repository, subject path, and revision fail closed;
- the existing MISSING -> source bridge -> read-only plan -> produced observation -> CURRENT loop remains intact.

## Public carrier contract

`rust_edit_class_observation` now takes both repository coordinates explicitly:

```text
REPO_ROOT CURRENT_REPOSITORY SUBJECT_REPOSITORY REVISION FILE
```

A hosted Oxc carrier can therefore bind `CURRENT_REPOSITORY` from the repository it intentionally checked out while preserving the selected subject coordinate separately.

## Boundary

- research/source adapter only;
- no second Rust syntax classifier;
- no remote-URL inference or GitHub-specific repository parser in the source core;
- current repository identity remains caller-supplied evidence and current Git head remains checkout-observed;
- no ownership, scheduling, merge, mutation, or action authority.

Supersedes the generic source API in #221 after hosted certification. #335 can compose on this repaired current-main source rather than the self-bound historical carrier.

### DIFF
--- a/examples/rust_edit_class_observation.rs
+++ b/examples/rust_edit_class_observation.rs
@@ -1,0 +1,63 @@
+use std::env;
+use std::error::Error;
+use std::path::{Path, PathBuf};
+
+#[allow(dead_code)]
+#[path = "../src/applicability.rs"]
+mod applicability;
+#[allow(dead_code)]
+#[path = "../src/discriminator_observation.rs"]
+mod discriminator_observation;
+#[allow(dead_code)]
+#[path = "../src/durable_obligation.rs"]
+mod durable_obligation;
+#[allow(dead_code)]
+#[path = "../src/evidence_planner.rs"]
+mod evidence_planner;
+#[allow(dead_code)]
+#[path = "../src/justification.rs"]
+mod justification;
+#[allow(dead_code)]
+#[path = "../src/observation_frontier.rs"]
+mod observation_frontier;
+#[allow(dead_code)]
+#[path = "../src/observation_probe_bridge.rs"]
+mod observation_probe_bridge;
+#[allow(dead_code)]
+#[path = "../src/rust_edit_class_source.rs"]
+mod rust_edit_class_source;
+
+use rust_edit_class_source::{RustEditClassSubject, collect_rust_edit_class_source};
+
+const USAGE: &str = "usage: rust_edit_class_observation REPO_ROOT CURRENT_REPOSITORY SUBJECT_REPOSITORY REVISION FILE";
+
+fn main() {
+    if let Err(error) = run() {
+        eprintln!("rust-edit-class-observation: {error}");
+        std::process::exit(1);
+    }
+}
+
+fn run() -> Result<(), Box<dyn Error>> {
+    let mut args = env::args().skip(1);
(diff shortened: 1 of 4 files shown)


teamleaderleo (commented):
Flint 🐺 exact-head review of `2e3a18a99667780e2d588fdebb0c6625aa6dbb37` on `main@26ff528e243808c8107d562aa55e8f6e0fe2d037`.

No code-correctness blocker found.

Attacks performed:
- subject repository/revision/exact Rust path remain the required evidence coordinate;
- current repository is independently caller-supplied and current revision is checkout-observed through `git rev-parse HEAD`; the source no longer copies subject repository into both sides of applicability;
- wrong current repository with identical subject revision/path retains the observed edit-class value but marks applicability INVALID;
- moved checkout head independently marks applicability INVALID while retaining the observed value;
- matching current repository/head/path produces APPLIES;
- root/non-single-parent and unchanged-anchor source states remain value UNKNOWN rather than manufacturing a partition;
- source receipt/applicability receipt preserve both subject and current coordinates;
- repository text, exact lowercase 40-hex revisions, and normalized repository-relative `.rs` paths fail closed on malformed input;
- the adapter reuses the existing #54 classifier and preserves the source-owned bridge/probe plus MISSING -> planned read-only probe -> produced observation -> CURRENT loop.

Explicit trust boundary: the source core treats `current_repository` as caller-supplied evidence and deliberately does not infer it from Git remotes. A hosted carrier must bind that input to the repository it intentionally checked out; this PR makes that dependency visible instead of self-binding it to the subject.

The branch was replayed byte-identically onto current main after #338 changed the repository CI gate. Hosted receipts on this exact replay: CI `32659920160` / job `97244396313` success and generated provenance `32659920157` success. Format, strict Clippy, population-aware live heads-up, full tests, and all repository/history/CI-filter/diff dogfood passed on GitHub-hosted `ubuntu-latest`.

Downstream note: #335 still uses the historical source CLI and must supply both `CURRENT_REPOSITORY` and `SUBJECT_REPOSITORY` after this lands; for its retained Oxc carrier both values are `oxc-project/oxc`.

PR #342, opened 2026-08-23
fix: keep paginated provider snapshots unknown

teamleaderleo:
Completes the reopened #292 boundary left after merged #338.

## Trigger

Merged #338 correctly routes GitHub provider facts through the shared #337 snapshot producer, but its final `read_current_provider_snapshot()` still hashes any successful `build_inventory_and_metadata()` result. That base reader may span multiple open-PR pages and additional per-PR file pages, so membership/head/path movement between those provider reads can create a mixed population that is still supplied as an exact current snapshot identity.

## Repair

Snapshot-bound active-work reads now use one strict provider acquisition function:

```text
one GitHub GraphQL response
  + pullRequests.hasNextPage == false
  + every PR files.hasNextPage == false
  -> exact population facts may be hashed

any required pagination
  -> exact provider snapshot unavailable
```

The rule applies to both axes:

- the frozen/required provider population used to bind the inventory;
- the independent current population reread used for `--current-provider-snapshot`.

If the **current** reread requires pagination, it is returned as unavailable, so #329 preserves provider-snapshot UNKNOWN. If the **required/frozen** read requires pagination, the advisory fails closed before minting a required identity from mixed reads.

The existing #333 terminal current-work reread remains separate and unchanged.

## Controls

The Python adapter controls now prove:

- an unpaginated population is built from exactly one GraphQL call;
- its source receipt explicitly names the single-read contract;
- top-level open-PR pagination rejects exact snapshot construction;
- per-PR file pagination rejects exact snapshot construction;
- the current-population wrapper converts either pagination case into unavailable/UNKNOWN;
- existing exact/moved/unavailable work-head and snapshot-equality quiet-prefilter controls remain intact.

## Boundary

This is deliberately conservative. It does not claim provider-wide transactional snapshot semantics across multiple calls. Repositories that exceed one `first: 100` PR page or one `files(first: 100)` page per PR lose exact population identity until a separately proven multi-read consistency protocol exists.

No hash changes, Rust applicability changes, TTL, checkout-HEAD inference, ownership, scheduling, merge, or mutation authority.

Issue #340, opened 2026-08-23
Prove GitHub active-work pagination is snapshot-consistent


teamleaderleo:
## Trigger

The #292 provider-population product lane is now complete through #329 (consumer applicability), #337 (canonical typed snapshot producer), and #338 (live GitHub second-population re-read). One provider-specific observation boundary remains UNKNOWN.

The live GitHub adapter reads each pull request's `headRefOid` in the repository-level `pullRequests` connection, while changed paths beyond the first page are fetched later through separate `pullRequest(number) { files(...) }` pagination calls.

A provider head can therefore move between the first PR-page observation and a later files-page observation. Repository code does not establish whether GitHub cursor semantics pin those separate calls to one immutable provider snapshot.

Potential mixed receipt:

```text
PR page observes work #N at head A
first 100 changed paths correspond to A
head advances to B
later files page is fetched after the advance
adapter emits head A + a path set that may include B-era evidence
```

The exact provider-snapshot hash can faithfully fingerprint that mixed fact set; hashing does not prove the read was internally atomic.

## Required distinction

Keep this separate from:

- provider population identity/re-query (#292/#338);
- repository/work-head applicability (#327/#333);
- per-item activity certainty (#289/#290).

Those axes can all be correct while one provider enumeration mixes coordinates internally.

## Smallest useful discriminator

Either:

1. authoritative GitHub documentation proving that the relevant GraphQL connections/cursors across these calls provide a stable snapshot for this use, or
2. a controlled GitHub-hosted replay that advances a PR head during multi-page changed-file enumeration and observes whether later pages remain pinned to the original head's path population.

If stability is established, retain the receipt and close this UNKNOWN. If mixed reads are reproduced, add the smallest fail-closed producer check/retry protocol that binds head and path evidence to one exact provider coordinate.

## Controls

- no wall-clock TTL;
- no repository `HEAD` substitute;
- no assumption that matching before/after head rereads proves intermediate file pages belong to one snapshot unless the counterexample supports that protocol;
- preserve full PR-membership pagination and explicit `updated_at desc` selection;
- keep provider-agnostic product snapshot identity unchanged unless evidence genuinely requires a new typed fact.

No ownership, scheduling, merge, or mutation authority.

Refs #292 #294 #298 #305 #308 #329 #333 #337 #338.

teamleaderleo:
Flint 🐺 documentation discriminator receipt.

[...]

PR #333, opened 2026-08-23
fix: bind active-work CI to provider-current head

teamleaderleo:
Closes #332 with the smallest live-adapter use of the merged #327 work-current gate.

## Change

The GitHub active-work CI adapter now performs one independent final GraphQL re-read of the current PR after building the frozen open-PR inventory.

It passes that re-read into the existing product through the exact #327 environment contract:

```text
CULTIST_REQUIRED_PROVIDER_REPOSITORY
CULTIST_CURRENT_PROVIDER_REPOSITORY
CULTIST_CURRENT_PROVIDER_WORK
CULTIST_CURRENT_PROVIDER_HEAD   # omitted when unavailable -> UNKNOWN
```

No checkout or Actions merge-view HEAD is used as provider work identity.

The existing cheap quiet prefilter remains only when the final provider-current repository/work/head exactly matches the frozen current inventory coordinate. A moved or unavailable final head escalates into product preflight even when frozen direct paths/coordination metadata looked quiet, allowing the merged gate to emit INVALID/UNKNOWN before a strong quiet/routing conclusion survives.

## Controls

The Python adapter controls cover:

- exact provider-current work response and exact-match prefilter admission;
- moved head -> prefilter denied;
- unavailable current PR/head -> prefilter denied;
- wrong repository/work -> prefilter denied;
- provider response with wrong work number fails closed;
- explicit product environment contains repository/work/head from provider context;
- an unavailable provider head removes any stale inherited `CULTIST_CURRENT_PROVIDER_HEAD` value.

The PR event itself provides the hosted integration replay: the live adapter will enumerate this PR and independently re-read its final provider head before invoking current merged product behavior when product analysis is needed.

## Boundary

- provider-current work axis only;
- no provider-population snapshot hash reimplementation in Python;
- #292's provider-population adapter/producer identity remains a separate follow-up;
- pagination consistency / mixed head-path reads remain UNKNOWN;
- no wall-clock TTL, checkout-HEAD inference, ownership, scheduling, merge, or mutation authority.

Validation authority is ordinary GitHub-hosted CI on repository-owned runners.

PR #345, opened 2026-08-23
fix: prove bounded file coverage in one provider read

teamleaderleo:
Refs #340. Follow-up to #342; does not revive cross-request pagination.

Current private Quarry dogfood #874 showed the landed exact-current producer correctly becomes unavailable when one open PR has >100 changed files. Quarry #729 currently has 102 files. Disposable capability probe #875 then proved that today's full 39-PR population can be covered in one GitHub GraphQL operation by requesting both bounded ends of each changed-file connection and checking their union against provider `totalCount`.

This patch admits that bounded case generically while preserving the #342 safety boundary:
- one GraphQL operation for the entire required/current population;
- top-level >100 open PRs still fails closed;
- each PR requests `files(first: 100)` and `files(last: 100)` in that same response;
- changed paths are admitted only when the unique union count exactly equals both provider `totalCount` values;
- count disagreement or incomplete union remains unavailable/UNKNOWN;
- no cursor follow-up requests are made by the exact snapshot path;
- #333 terminal current-work head reread and per-item activity semantics remain unchanged.

Controls added for 1-file exact coverage, 102-file overlapping-slice exact coverage, top-level PR pagination refusal, 201-file incomplete-union refusal, and first/last count disagreement refusal.

Provider capability receipt from Quarry #875: hosted run/job `32663136113 / 97252216527`, 39 open PRs, #729 102/102 unique paths, receipt SHA-256 `c96e31ccca92496cadd1f8b51716783e832892887f4b81d0f2401661d7b7941d`, artifact `9499264449`.

Base `81cb89864a01cc2254c7744e9bd6518425e64458`; head `f63ac0da19cbef573b17268169acae510c0a65f6`. Hosted CI required; author review does not count as independent review.

### DIFF
--- a/scripts/active_work_heads_up_snapshot_ci.py
+++ b/scripts/active_work_heads_up_snapshot_ci.py
@@ -15,6 +15,75 @@ import active_work_heads_up_ci as base
 from github_provider_snapshot_ci import derive_provider_snapshot_identities
 
 
+SINGLE_READ_PR_PAGE_QUERY = r"""
+query($owner: String!, $name: String!) {
+  repository(owner: $owner, name: $name) {
+    pullRequests(
+      states: OPEN
+      first: 100
+      orderBy: {field: UPDATED_AT, direction: DESC}
+    ) {
+      nodes {
+        number
+        title
+        url
+        body
+        headRefName
+        headRefOid
+        updatedAt
+        isDraft
+        filesFirst: files(first: 100) {
+          totalCount
+          nodes { path }
+        }
+        filesLast: files(last: 100) {
+          totalCount
+          nodes { path }
+        }
+      }
+      pageInfo { hasNextPage endCursor }
+    }
+  }
+}
+"""
+
+
+def complete_changed_paths_from_single_response(node: dict[str, object]) -> list[str]:
+    """Return exact changed paths only when one response proves completeness."""
+    number = int(node["number"])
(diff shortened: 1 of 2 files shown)


teamleaderleo (commented):
Rook 🦉 review of exact head `f63ac0da19cbef573b17268169acae510c0a65f6`: no semantic blocker found.

The repair preserves the #342 safety boundary while recovering a bounded utility case:
- one provider GraphQL operation only;
- top-level PR continuation still fails closed;
- each work item gets `files(first: 100)` + `files(last: 100)` in that same response;
- both provider counts must be readable and equal;
- the unique union must equal that count before paths are admitted;
- incomplete coverage and count disagreement remain unavailable/UNKNOWN;
- no cursor follow-up path is reintroduced;
- the terminal current-work reread remains a separate coordinate.

I also traced the downstream helpers: `work_item` receives a normalized synthetic `files` connection, while `metadata_item` only consumes id/body/head/update metadata, so the raw `filesFirst/filesLast` node does not create a hidden compatibility break.

Exact-head hosted CI `32663509825` and provenance `32663509809` are green on base `81cb89864a01cc2254c7744e9bd6518425e64458`.

Retained limit: this earns bounded file-set coverage under the existing one-response contract; it does not prove cross-request pagination consistency, admit top-level >100 PR populations, or admit file populations whose first/last bounded union cannot prove completeness.

PR #344, opened 2026-08-23
research: replay refinement candidate readiness on current main

teamleaderleo:
Progresses #240 as the current-main successor to historical #243, now on the merged #335 selected-observation loop.

## Retained contract

The candidate-readiness evaluator preserves two independent axes for every retained refinement candidate:

```text
replay_status / replay_result
current exact evidence readiness
```

`evidence_status = current` only when every discriminator has an exact mapping and every mapped frontier is CURRENT. Replay status remains unchanged.

## Current-main fixture repair

Current main deliberately keeps the focused Oxc edit-class receipt in the dedicated fixture:

`research/refinement-observation-requirements/oxc-focused-edit-class-v1.json`

rather than the broad discriminator corpus. This replay composes that focused receipt into the request exactly as merged #287/#335 do, while keeping the #243 evaluator semantics unchanged.

Controls preserve the asymmetric cases:

- selected `syntax-changing-current-cohort`: `weakened + current`;
- remove the exact focused receipt and add a wrong-path edit-class observation: replay remains `weakened`, evidence becomes blocked with an exact MISSING frontier;
- remove the selected candidate's D@S mapping: replay remains `weakened`, evidence stays blocked with `missing_requirement_mappings=[edit_class]`;
- provide explicit current evidence for `reverse-edit-class-control`: replay remains `rejected_no_improvement` while evidence is current;
- provide explicit current evidence for `singleton-commit-partition`: replay remains `rejected_overfit` while evidence is current;
- retained rejected candidates without mappings remain explicitly evidence-blocked;
- every candidate preserves its exact replay receipt.

## Boundary

Descriptive research layer only. No candidate promotion, automatic selection, ranking score, source acquisition, ownership, scheduling, merge, effect, or mutation authority. Exact subject/currentness semantics remain owned by the merged requirement/frontier layers.

Supersedes #243 after hosted certification. Refs #148 #179 #185 #210 #216 #220 #227 #231 #234 #240 #243 #335.

Issue #278, opened 2026-08-19
Research promotion receipt reuse instead of exact-main churn


teamleaderleo:
Dogfood from the long 2026-08-19 promotion sweep exposed a recurring concurrency cost: successful merge-view CI was repeatedly discarded whenever unrelated `main` commits landed, causing byte-identical research children to be reanchored and rerun several times.

The repository now has prose guidance in `docs/concurrent-work-promotion.md`. This issue asks whether Cultist can make the decision deterministic enough to surface before a worker starts another full CI cycle.

## Candidate input

```text
PR/head identity
successful tested head/tree/merge-view identity
current PR merge view
intervening main commits + changed paths
caller-supplied known consumed contract paths when needed
```

## Candidate output

Keep the result evidence-bearing instead of a binary safety oracle:

```text
receipt_reusable
  proven identical effective merge tree / equivalent tested repository state

rerun_required
  direct changed-path overlap, merge conflict, changed applicable policy/CI gate,
  or another proven compatibility input changed

inspect_semantic_overlap
  path-disjoint but semantic independence remains UNKNOWN
```

The first version should never infer semantic independence from path disjointness alone.

## Acceptance controls

1. Same tested/current merge-tree identity -> `receipt_reusable`.
2. Branch commit metadata changes with identical tree and relevant base semantics -> receipt can transfer.
3. Intervening main commit touches a branch path -> `rerun_required`.
4. Intervening main commit changes a directly declared consumed contract path -> `rerun_required`.
5. Disjoint paths with no stronger evidence -> `inspect_semantic_overlap`, not automatic rerun or automatic safety.
6. GitHub reports merge conflict -> `rerun_required` / reconcile first.
7. Output preserves exact SHAs/tree IDs/commit range and the reason for the disposition.

## Dogfood target

Replay the repeated reanchors from the #194 / #201 / #156 promotion sequence and measure how many full CI reruns the first bounded policy would have avoided while still flagging the known zero-path-overlap project-memory contract collision family as semantically UNKNOWN.

Research first. Promotion policy should remain conservative until the negative controls are convincing.

Refs #96 #137 #156 #194 #201.

teamleaderleo:
First complete historical dogfood result from #194:

Early successful merge-view run `32249440070` checked branch head `b54ed3213994c96ec818ef36bb9728b0dc1f7eb6` against real base `a6026e084ae6f1d88e8d7f2322473872f82c3278`.

[...]

PR #346, opened 2026-08-23
research: dogfood promotion policy overlap

teamleaderleo:
Progresses #278 with one real current Cultist promotion episode.

## Real receipt

#344 first passed hosted CI on exact head `12fc7ea0b031e7128d0d72cc861f2d1ad78b4e30` against base `c84e151714d3d9a53a0723dbbf384e0df7e12242`.

After #342 landed, the #344 semantic payload remained byte-identical across the same four path/blob pairs, but the applicable CI execution input changed: #342 modified the active-work snapshot adapter and its control script, both executed by the ordinary CI gate.

#344 was therefore replayed on exact head `18c204c39c38f60e43d25e9ddc2a325e07ac7f0a` against base `81cb89864a01cc2254c7744e9bd6518425e64458`, and the new hosted run passed.

## Carrier

This test feeds the exact old/current head, base-tree, merge-tree, and branch blob coordinates into the merged #278 evaluator. The branch change-set fingerprint is computed through the existing production helper and is identical across the two states.

With the two executed active-work scripts declared as applicable policy, the one intervening #342 commit must classify as:

```text
rerun_required
reason = applicable_policy_overlap
```

The paired negative control removes that policy declaration while keeping compatibility scope incomplete. The same byte-identical payload must remain:

```text
inspect_semantic_overlap
reason = semantic_independence_unknown
```

It must not become receipt-reusable merely because branch paths are disjoint.

## Boundary

- test-only dogfood of the existing promotion evaluator;
- no new promotion state or policy rule;
- no timestamp/freshness heuristic;
- no claim that every main movement requires rerun;
- exact executed CI inputs remain part of compatibility applicability when they change.

### DIFF
--- a/tests/promotion_receipt_policy_overlap_dogfood.rs
+++ b/tests/promotion_receipt_policy_overlap_dogfood.rs
@@ -1,0 +1,172 @@
+#[allow(dead_code)]
+#[path = "../src/promotion_change_set.rs"]
+mod promotion_change_set;
+#[allow(dead_code)]
+#[path = "../src/promotion_receipt.rs"]
+mod promotion_receipt;
+
+use promotion_change_set::{PromotionChangeSetEntry, fingerprint_promotion_change_set};
+use promotion_receipt::{
+    CurrentPromotionState, InterveningCommit, PROMOTION_RECEIPT_SCHEMA_VERSION,
+    PromotionPathOverlapKind, PromotionReceiptDisposition, PromotionReceiptReason,
+    PromotionReceiptRequest, TestedPromotionState, evaluate_promotion_receipt,
+};
+
+const OLD_HEAD: &str = "12fc7ea0b031e7128d0d72cc861f2d1ad78b4e30";
+const OLD_TREE: &str = "1da2cc89d48b81bf3b9e10696651de9a4006e1d7";
+const OLD_BASE: &str = "c84e151714d3d9a53a0723dbbf384e0df7e12242";
+const OLD_BASE_TREE: &str = "28e49529d627463c8fa2eef550c1970cede5c1ac";
+const OLD_MERGE_TREE: &str = "1da2cc89d48b81bf3b9e10696651de9a4006e1d7";
+
+const CURRENT_HEAD: &str = "18c204c39c38f60e43d25e9ddc2a325e07ac7f0a";
+const CURRENT_TREE: &str = "d3942887c7f9b8ecfa3b6f4fe0cf46213eda08aa";
+const CURRENT_BASE: &str = "81cb89864a01cc2254c7744e9bd6518425e64458";
+const CURRENT_BASE_TREE: &str = "99513f270a58d0775c6ff7dc8752fdb7999a049f";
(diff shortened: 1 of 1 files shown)


teamleaderleo (commented):
Rook 🦉 self-review of exact head `23f17f050e512840f72da11baf0fef144764b683`: no semantic blocker found.

This dogfoods the merged promotion evaluator on the exact #344 rerun from this sweep:
- old and current branch payloads use the same four path/blob pairs, and the shared production change-set fingerprint is equal;
- old/current base trees and effective merge trees differ;
- the single intervening #342 commit changed the two active-work scripts executed by ordinary CI;
- declaring those scripts as applicable policy yields `rerun_required + applicable_policy_overlap` with both overlaps retained;
- the paired control removes the policy declaration while keeping compatibility scope incomplete and correctly stays `inspect_semantic_overlap + semantic_independence_unknown`.

So the receipt does not overgeneralize from `main` movement or branch-path disjointness. It requires a rerun because an executed policy input changed, and preserves UNKNOWN when that applicability declaration is absent.

Exact-head hosted CI `32667058117` / run #2780 passed every gate; provenance `32667058119` / run #685 passed. Author review does not count as independent review.

teamleaderleo (commented):
Flint 🐺 independent review of exact head `23f17f050e512840f72da11baf0fef144764b683`.

No code-correctness blocker found in the one-file dogfood carrier.

I attacked the historical coordinates and the negative control rather than accepting the rerun story from chronology alone. The carrier uses the real #344 old/current payload as the same four path/blob pairs and computes equality through the existing production change-set fingerprint. The old base is exactly #342's parent `c84e151714d3d9a53a0723dbbf384e0df7e12242`; current base `81cb89864a01cc2254c7744e9bd6518425e64458` is #342's merge commit, whose product diff changes the two active-work adapter/control scripts named by the test and executed by ordinary CI. Declaring those exact executed inputs as applicable policy therefore produces `rerun_required + applicable_policy_overlap` without using main movement as the discriminator.

The paired control is important and holds: removing the applicable-policy declaration while keeping compatibility scope incomplete yields `inspect_semantic_overlap + semantic_independence_unknown`; byte-identical branch payload and disjoint branch paths do not become proof of reusable CI evidence.

Hosted validation is green on this exact head: ordinary CI `32667058117` and generated provenance `32667058119` both succeeded. This COMMENT records code/evidence review only; it grants no external approval authority.

teamleaderleo (commented):
Flint 🐺 current-head review of `d5007dd830ac732e1e7ce4172d09d5b6d14ee3f7` after replay on `main@b79ea26f40cec8c39666241aa23b39c38b41e772`.

No code-correctness blocker found. The one test blob is byte-identical to the earlier reviewed carrier. Revalidation was required because merged #345 changed the active-work adapter/control executed by ordinary CI after the earlier green receipt; that is exactly the applicable-CI-input condition this dogfood is meant to preserve. Current hosted CI `32739828925` passed format, Clippy, active-work controls/live heads-up, full tests, and all dogfood; generated provenance `32739829008` passed.

The semantic result remains narrow: exact executed policy overlap -> rerun; absent a complete applicability declaration -> semantic independence stays UNKNOWN.

PR #201, opened 2026-08-19
research: prove KNOWN can outrun current applicability

teamleaderleo:
Closes the #195 discriminator with an explicit negative control over the now-landed #194 frontier.

## Proven incoherent state

The test composes the real modules:

```text
applicability.rs
+ discriminator_observation.rs
+ observation_frontier.rs
```

and shows the v1 types permit a same-subject KNOWN value to outrun current applicability.

### Moved head

```text
required revision = old-head
current revision  = new-head
shared applicability = INVALID
frontier             = CURRENT
```

### Missing current revision

```text
required revision = exact-head
current revision  = missing
shared applicability = UNKNOWN
frontier             = CURRENT
```

The observation's `applicability_ref` is only an opaque receipt pointer in v1, so #194 classifies currentness from `value_state=KNOWN` alone.

This commit deliberately keeps that behavior green. It is the proof artifact #210 must close; no production or research implementation semantics change here.

## Current-main promotion receipt

The exact 101-line counterexample blob was reanchored twice as unrelated main work landed during earlier successful runs. Final current base and head:

```text
main: 79c5c64d2cd17869715c0818e0bf86bdad5b7322
head: 3cf9090dfb474adaac6ab773c357627c37c3f9e6
commits: 1
files: 1
```

Exact-current ordinary CI:

```text
run: 32270569205
job: 96125697857
result: success
```

Generated-provenance review:

```text
run: 32270569233
job: 96125697877
result: success
```

The modern external-reference post-write detector, active-work preflight, full tests, and repository/history/CI-filter/diff dogfood all passed. Main remained at the exact base through the final gate.

## Meaning

The green result proves the forbidden composition exists in v1:

```text
KNOWN old value
+ INVALID | UNKNOWN current applicability
-> Current frontier
```

#210 chooses Model B and separates value knowledge from typed applicability so this state can no longer suppress acquisition.

Files: `tests/known_stale_observation_frontier.rs` only.

Refs #123 #145 #185 #187 #190 #194 #195 #210.

PR #347, opened 2026-08-23
research: dogfood divergent promotion policy overlap

teamleaderleo:
Progresses #278 with the historical #201 promotion episode named in the existing dogfood handoff.

## Historical coordinates

Early successful #201 state:

```text
head       1614cc2ae82df50ec3c8b5c4a9e428ad01c1d50f
tree       a90a3317c50d5d7d693b948cc9414315056c628f
base       d3482178aee3cba7af496abd3adab8ac162639cb
base tree  2206ff89408e902354494cac853e0311c2143f57
CI         32250242114
```

Final promoted state:

```text
head       3cf9090dfb474adaac6ab773c357627c37c3f9e6
tree       889e34a998fe268986718bf21e72263503a1a05b
base       79c5c64d2cd17869715c0818e0bf86bdad5b7322
base tree  cb9fdb279cdff8fdf6362d317166f7d82df51e87
CI         32270569205
```

The bases diverged from merge base `c2133131038f33a98f7bc7d206ca6e4284be420a`, while the exact one-file branch payload blob stayed unchanged.

## Carrier

The test uses the repository's full Git history to derive both base-only commit counts and complete changed-path receipts at execution time, then supplies them to the merged divergence-aware promotion evaluator.

Compatibility objects are the three modules the #201 test consumes plus the ordinary CI workflow:

- `src/applicability.rs`;
- `src/discriminator_observation.rs`;
- `src/observation_frontier.rs`;
- `.github/workflows/ci.yml` as applicable policy.

The three consumed contract blob IDs are identical between tested and final bases. The CI workflow blob changes.

Expected result:

```text
base_relation = diverged
disposition = rerun_required
reason = applicable_policy_overlap
```

Paired control makes only the policy object artificially equal while preserving the complete divergent range receipts and incomplete semantic scope. That must remain `inspect_semantic_overlap / semantic_independence_unknown`, proving divergence alone is not a rerun rule.

## Boundary

- test-only historical dogfood;
- no new promotion taxonomy or policy rule;
- no timestamp heuristic;
- exact Git object/range receipts are derived under hosted CI rather than hand-truncated;
- no claim that divergent bases alone require rerun.

### DIFF
--- a/tests/promotion_base_lineage_pr201_dogfood.rs
+++ b/tests/promotion_base_lineage_pr201_dogfood.rs
@@ -1,0 +1,239 @@
+#[allow(dead_code)]
+#[path = "../src/promotion_base_lineage.rs"]
+mod promotion_base_lineage;
+#[allow(dead_code)]
+#[path = "../src/promotion_change_set.rs"]
+mod promotion_change_set;
+#[allow(dead_code)]
+#[path = "../src/promotion_receipt.rs"]
+mod promotion_receipt;
+
+use std::process::Command;
+
+use promotion_base_lineage::{
+    PROMOTION_BASE_LINEAGE_SCHEMA_VERSION, PromotionBaseLineageRequest, PromotionBaseRange,
+    PromotionBaseRelation, PromotionCompatibilityObjectKind, PromotionCompatibilityObjectState,
+    evaluate_promotion_base_lineage,
+};
+use promotion_change_set::{PromotionChangeSetEntry, fingerprint_promotion_change_set};
+use promotion_receipt::{
+    CurrentPromotionState, PROMOTION_RECEIPT_SCHEMA_VERSION, PromotionReceiptDisposition,
+    PromotionReceiptReason, PromotionReceiptRequest, TestedPromotionState,
+};
+
+const MERGE_BASE: &str = "c2133131038f33a98f7bc7d206ca6e4284be420a";
+const TESTED_BASE: &str = "d3482178aee3cba7af496abd3adab8ac162639cb";
+const TESTED_BASE_TREE: &str = "2206ff89408e902354494cac853e0311c2143f57";
+const TESTED_HEAD: &str = "1614cc2ae82df50ec3c8b5c4a9e428ad01c1d50f";
+const TESTED_TREE: &str = "a90a3317c50d5d7d693b948cc9414315056c628f";
+const TESTED_CHECK: &str = "github-actions:ci/32250242114";
+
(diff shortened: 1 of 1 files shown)


teamleaderleo (commented):
Rook 🦉 self-review of exact head `06f5b3117a971c8e4b6eeed9a0b7fd0e31e04027`: no semantic blocker found.

The hosted carrier resolves #201's historical Git objects/ranges directly under `fetch-depth: 0` and proves:
- early/final one-file branch payload uses the same exact blob and production change-set fingerprint;
- bases are genuinely divergent from `c2133131038f33a98f7bc7d206ca6e4284be420a`, with complete tested-only/current-only changed-path receipts derived at execution time;
- `src/applicability.rs`, `src/discriminator_observation.rs`, and `src/observation_frontier.rs` have identical tested/current object IDs;
- `.github/workflows/ci.yml` changes between tested/final bases;
- the divergence-aware evaluator therefore returns `rerun_required + applicable_policy_overlap`, with no branch-path or consumed-contract reason;
- paired control makes only the policy object equal and correctly preserves `inspect_semantic_overlap + semantic_independence_unknown`, so divergence alone is not treated as rerun authority.

Exact-head CI `32667848613` / run #2785 and provenance `32667848584` / run #687 are green. Author review does not count as independent review.

teamleaderleo (commented):
Flint 🐺 independent/current-head review of `9e5aae7fa9bc65dabcb1ef2eff5d9633560596a9`.

No code-correctness blocker found. I attacked the historical coordinates and the divergent-base negative control. The test derives both base-only histories from full Git history at execution time, verifies the one-file #201 branch blob is byte-identical at tested/current heads, and verifies the three consumed contract objects (`applicability`, `discriminator_observation`, `observation_frontier`) are byte-identical across the divergent bases. The only declared compatibility object that moves is `.github/workflows/ci.yml`, so `Diverged + ApplicablePolicyOverlap -> RerunRequired` is supported by an exact changed execution input rather than by divergence itself.

The paired control changes only the policy object receipt to equal while retaining the real divergent range receipts and incomplete semantic scope; it stays `InspectSemanticOverlap + SemanticIndependenceUnknown`, which prevents base divergence or branch-path disjointness from manufacturing receipt reuse/rerun semantics.

Hosted validation on this exact replay is green: CI `32739984928` and generated provenance `32739984893` succeeded. The replay base was `b79ea26f40cec8c39666241aa23b39c38b41e772`; subsequent main `f4755bf40767164e9d7302898ed7f0e5a60f7b3a` adds only `tests/promotion_receipt_policy_overlap_dogfood.rs`, with no consumed-contract or applicable-CI-policy movement for this carrier. No further rerun is justified by that one intervening commit.

PR #349, opened 2026-08-24
research: dogfood reusable promotion receipt

teamleaderleo:
Progresses #278 with the missing real positive receipt-reuse case from historical #156.

## Historical coordinate

#156 final promotion authority was exact head `e33cc0e736d3d9092cd23782cef4a8ad2cd1935a`, with hosted CI `32272350023` and provenance `32272350050`.

A minute later the branch moved to `9299022e76406617c06858d9d6d441c7e13b43b5` and ordinary CI `32272437964` plus provenance `32272438358` ran again.

GitHub's commit API still resolves both heads to the same exact parent/base `eb9c0eeed395b9e40addc30c408d4ff83f24ab42`, the same full Git tree `1ac85baec7efa0ca5170bdcf54c206406dccc60c`, and the same four justification blob identities.

## Hosted counterexample that tightened the carrier

CI `32742251730` reached the historical test and failed because the hosted `actions/checkout@v4` clone, despite `fetch-depth: 0`, no longer contained the older tested head object `e33cc0e...` (`fatal: bad object`). GitHub's commit API still resolves that object and its exact tree/base/blob coordinates.

That failure is retained as a useful boundary: a full checkout of currently reachable refs is not a durable archive of arbitrary old PR heads. Promotion receipt reuse therefore must consume retained tested coordinates rather than assume they can always be reconstructed later from checkout history.

## Carrier

The single test now treats the exact API-verified #156 coordinates as the retained historical receipt, fingerprints the four retained branch blobs through the existing production `promotion_change_set` helper, and evaluates the existing #284 policy.

Expected result:

```text
head changed
retained base unchanged
retained full tree unchanged
retained effective merge tree unchanged
branch change-set unchanged
-> receipt_reusable
   exact_effective_merge_tree_identity
```

This is a real case where #278 could have avoided one full ordinary CI rerun. It complements #346/#347, which correctly classify real CI-policy movement as rerun-required.

## Boundary

- test-only historical dogfood;
- no new promotion disposition or rule;
- no semantic-independence inference from path disjointness;
- no timestamp heuristic;
- exact retained tree/base/payload identity is the reason for reuse;
- checkout reachability is not treated as receipt validity;
- no automated rebase, merge, rerun, or mutation action.

### DIFF
--- a/tests/promotion_receipt_pr156_reuse_dogfood.rs
+++ b/tests/promotion_receipt_pr156_reuse_dogfood.rs
@@ -24,7 +24,12 @@ fn git(args: &[&str]) -> String {
         .args(args)
         .output()
         .expect("git must execute in hosted CI");
-    assert!(output.status.success(), "git {:?} failed: {}", args, String::from_utf8_lossy(&output.stderr));
+    assert!(
+        output.status.success(),
+        "git {:?} failed: {}",
+        args,
+        String::from_utf8_lossy(&output.stderr)
+    );
     String::from_utf8(output.stdout)
         .expect("git output must be UTF-8")
         .trim()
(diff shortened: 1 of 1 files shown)


teamleaderleo (commented):
Flint 🐺 independent/current-head review of exact head `5842b9ec6fe33bbabc025d861387c649e8d91f83` on `main@d698aee145ba10896990c68e7b9ae70b35f5e9ef`.

No code-correctness blocker found.

The positive historical reuse case now tests the right contract: retained exact promotion coordinates, not later checkout recoverability. GitHub's commit API independently confirms historical #156 heads `e33cc0e736d3d9092cd23782cef4a8ad2cd1935a` and `9299022e76406617c06858d9d6d441c7e13b43b5` share parent `eb9c0eeed395b9e40addc30c408d4ff83f24ab42`, tree `1ac85baec7efa0ca5170bdcf54c206406dccc60c`, and the four retained justification blobs. The test feeds those retained tree/base/blob coordinates plus the successful tested check refs into the existing promotion evaluator; the branch change-set fingerprint is invariant to entry order and the evaluator returns `ReceiptReusable` with `ExactEffectiveMergeTreeIdentity`.

The failed hosted attempt `32742251730` is an important retained negative control: `actions/checkout@v4` with `fetch-depth: 0` could not resolve the older orphaned PR head (`fatal: bad object`) even though GitHub's commit API still resolves it. A durable promotion receipt therefore cannot assume arbitrary old PR heads remain reconstructible from currently reachable checkout refs.

Hosted validation on the repaired exact head is green: CI `32742751033` passed format, strict Clippy, current active-work controls/live advisory, full tests, and all dogfood; generated provenance `32742752003` passed. Main did not move between this run and review.

Historical measurement: the later #156 run `32272437964` repeated ordinary CI after the tested `32272350023` state even though the retained effective repository state was identical, so this first bounded policy would have avoided one full CI rerun in that episode.

PR #255, opened 2026-08-19
research: plan only admitted refinement acquisition frontiers

teamleaderleo:
Progresses #190 as a pure composition child of green #251.

## Gate before planning

#251 separates candidate dispositions:

```text
replay_rejected
unselected
satisfied
requirement_mapping_needed
observation_acquisition_needed
```

This carrier feeds only `observation_acquisition_needed.acquisition_frontiers[]` into #216. Every other disposition carries zero planner frontiers.

## Held-out replay visibility

Strengthened #251 defines survival as status-level non-rejection while preserving the complete `ReplayResult`, including held-out `passed | not_run | unknown`.

This child now proves the planner consumes that contract without silently strengthening it:

```text
selected survivor
held_out_status = not_run
exact mapped observation missing
-> observation_acquisition_needed
-> held_out not_run remains visible on the authorizing disposition
-> exact #221 source probe plans successfully
```

So current policy permits evidence acquisition for a selected status survivor even while held-out replay is incomplete. Any future stricter held-out prerequisite must become an explicit gate from the preserved replay result.

## Standard CI

A temporary local Git history runs through the real #221 Rust edit-class source adapter. The retained selected Oxc replay candidate is explicitly rebound to the test subject through #231 mapping evidence.

Missing exact observation:

```text
selected survivor
-> observation_acquisition_needed
-> exact MISSING frontier
-> #221 bridge/probe
-> #216 planned
-> #145 evidence plan selected
```

After the source observation is inserted, the same selected candidate becomes `satisfied` with zero acquisition frontiers. Exact mapped UNKNOWN and INVALID frontiers also reach the existing planner while preserving their frontier status.

Current, missing-mapping, replay-rejected, and unselected dispositions create zero planner inputs.

## Public Oxc carrier

The dedicated workflow begins from retained #179/#231/#185 data, withholds exact `edit_class@228e8e0f...:rules.rs`, and preserves wrong-path + pinned-head UNKNOWN controls. #251 admits exactly one Oxc acquisition frontier and keeps both replay-rejected alternatives quiet.

Only that admitted frontier is passed to the #221 source bridge/probe and #216 planner. The workflow inserts the produced `syntax_changed` observation and #251 changes the selected candidate from `observation_acquisition_needed` to `satisfied` while rejected alternatives remain `replay_rejected`.

## Latest validation

Current head:

```text
1065086c45ad167037d290b83d77d52975d1f1a9
```

Ordinary CI:

```text
32274596102 success
```

[...]

PR #251, opened 2026-08-19
research: gate refinement investigation demand

teamleaderleo:
Progresses #248 as the replay/selection/currentness gate before expensive refinement investigation.

## Core disposition

For each retained replay candidate:

```text
replay rejected
  -> replay_rejected

survives but is not the selected transition
  -> unselected

selected survivor + all mapped exact observations current
  -> satisfied

selected survivor + missing requirement mapping
  -> requirement_mapping_needed

selected survivor + mapped MISSING | UNKNOWN | INVALID observation frontier
  -> observation_acquisition_needed
```

Rejected and unselected candidates emit zero acquisition frontiers.

## Held-out replay contract

The parent refinement episode permits kept candidates with held-out status:

```text
passed | not_run | unknown
```

and rejects kept candidates whose held-out replay is `failed`.

This lane therefore defines **survival as status-level non-rejection**. Held-out completeness remains an independent receipt axis and is never rewritten to `passed`.

`RefinementInvestigationDisposition` now carries the complete source `ReplayResult` in addition to the convenience `replay_status`, preserving expected/lost cases, counterexample counts, and `held_out_status` for downstream readers.

New executable controls prove:

```text
selected survivor + held_out unknown + current evidence
  -> satisfied
  -> held_out unknown preserved

selected survivor + held_out not_run + exact mapped observation missing
  -> observation_acquisition_needed
  -> held_out not_run preserved
```

A downstream planner may choose a stricter held-out prerequisite later, but it must do so explicitly from this preserved replay result rather than inferring completion from candidate status.

## Current validation

Latest semantic head includes the held-out visibility change and its dedicated contract test. Ordinary CI run `32273560605` passed rustfmt, strict Clippy, the complete test suite, active-work heads-up, all repository/history/CI-test dogfood, and diff text + JSON.

The focused note is retained in:

```text
research/refinement-investigation-held-out-contract.md
```

No acquisition executes here; this layer only says where investigation is justified.

Refs #179 #190 #231 #243 #248 #255.

PR #154, opened 2026-08-19
research: prove fingerprint-bound expansion of terse refs

teamleaderleo:
Progresses #119 / #115 using the now-merged exact report fingerprint.

## Question

Can terse output actually keep provenance expandable on demand without pretending report-local refs are durable IDs?

## Experiment

A research integration fixture builds one `AnalysisReport` with:

- report-level UNKNOWN;
- finding `F1`;
- claim `F1.C1`;
- two support receipts under that claim;
- one question.

The current terse renderer omits the support receipts.

The fixture then resolves report-local references only when paired with the exact C1/SHA-256 report fingerprint:

```text
C1          top-level claim
F1          finding
F1.C1       finding claim
F1.C1.E2    second evidence receipt
F1.Q        question
```

## Controls

- terse omits the exact-head support receipts;
- `F1.C1` expansion recovers the claim and both receipts;
- `F1.C1.E2` recovers one exact receipt directly;
- report-level claim and question expand from the same snapshot;
- mutating claim semantics invalidates the old fingerprint before resolving `F1.C1`;
- pure finding reorder invalidates the old fingerprint before positional `F1` lookup;
- malformed and out-of-range refs fail explicitly.

## Meaning if green

This earns a narrow expansion contract without durable semantic IDs:

```text
exact report fingerprint + report-local reference
    -> safe expansion inside that immutable snapshot
```

It composes #128's positional-ref counterexample with #131's exact report fingerprint.

It does **not** solve semantic lineage across reports, renames, or changed findings.

## Boundary

- test-only resolver; no public CLI/protocol syntax yet;
- no persistent `F1` identity;
- no report store/service;
- no authority/disposition semantics;
- the containing envelope still needs #134-style ambient context binding where revision/task/provider coordinates are required.

Refs #115 #119 #121 #127 #128 #131 #134.

PR #351, opened 2026-08-24
history: stream git output and bound retained changed paths

teamleaderleo:
## Summary

This advances a bounded slice of #51 (the history half). The historical-companions experiment now consumes the batched `git log` output as a stream, and individual changed paths stop being retained once a commit crosses the broad-commit threshold. It does not close the whole issue.

## Changes

- `src/history.rs` spawns Git with piped stdout/stderr (stdin nulled), consumes stdout incrementally through a buffered line reader, and drains stderr on a separate thread so the child cannot block on a full pipe while stdout is read. Nonzero status still reports the captured stderr verbatim; parse failures terminate the child safely.
- The streaming record parser preserves the previous grammar (`\x1e` records, `\x1f` metadata fields, per-record path lines) with identical failure behavior for malformed or truncated records.
- While a commit's deduplicated path set is at or below `broad_commit_threshold`, paths are retained exactly as before (same `BTreeSet` dedupe and ordering). On the insertion that first exceeds the threshold, retained path strings are dropped and parsing switches to counting only; the reported count stays exact because `git log --name-only --no-renames` lists each changed path once per commit.
- `HistoricalCommit.changed_paths` is now the authoritative count used by revert/broad filtering and `ExcludedCommit`. Report schema, commit order, companion ordering, revert handling, root/merge/path-filter behavior, and error surfaces are unchanged.

## Evidence

- End-to-end broad-commit exactness: a 6-path commit against threshold 2 reports `changed_paths == 6`; this repository's own dogfood run counts the mass replay commit at exactly 349 paths while retaining none of them.
- Unit tests prove retained storage empties after crossing while counting continues, threshold-boundary commits stay fully retained, duplicates below the threshold still dedupe exactly, blank/prelude whitespace is ignored, and malformed/truncated fixtures fail safely without panicking.
- `performance::capture` instrumentation proves history extraction still uses exactly one Git subprocess.
- Verification: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, all focused history tests, and the full local suite pass except `rust_edit_class_source_current_main::focused_source_values_and_focus_admission_survive_current_main`, which fails identically at base 306423d (`pathspec 'master'` environment dependency, unrelated to this change).

🌙 Nocturne

### DIFF
--- a/src/history.rs
+++ b/src/history.rs
@@ -1,10 +1,12 @@
 use std::collections::{BTreeMap, BTreeSet};
 use std::error::Error;
 use std::fs::File;
-use std::io::Read;
+use std::io::{BufRead, BufReader, Read};
 use std::path::{Path, PathBuf};
 #[cfg(test)]
 use std::process::Command;
+use std::process::{Child, Stdio};
+use std::thread;
 
 use serde::Serialize;
 
@@ -93,6 +95,7 @@ pub struct HistoryReport {
 struct HistoricalCommit {
     summary: CommitSummary,
     paths: Vec<PathBuf>,
+    changed_paths: usize,
 }
 
 pub fn analyze_historical_companions(
@@ -110,7 +113,7 @@ pub fn analyze_historical_companions(
         return Err("history broad-commit threshold must be greater than zero".into());
     }
 
-    let commits = read_anchor_history(root, anchor, options.max_commits)?;
+    let commits = read_anchor_history(root, anchor, options)?;
     let discovered_commits = commits.len();
     let mut considered = Vec::new();
     let mut excluded_commits = Vec::new();
@@ -120,21 +123,22 @@ pub fn analyze_historical_companions(
             excluded_commits.push(ExcludedCommit {
                 commit: commit.summary,
                 reason: "revert commit".to_string(),
-                changed_paths: commit.paths.len(),
+                changed_paths: commit.changed_paths,
             });
             continue;
         }
-        if commit.paths.len() > options.max_paths_per_commit {
+        if commit.changed_paths > options.max_paths_per_commit {
             excluded_commits.push(ExcludedCommit {
                 commit: commit.summary,
                 reason: format!(
                     "broad commit changed more than {} paths",
                     options.max_paths_per_commit
                 ),
-                changed_paths: commit.paths.len(),
+                changed_paths: commit.changed_paths,
             });
             continue;
         }
+        debug_assert_eq!(commit.changed_paths, commit.paths.len());
         considered.push(commit);
     }
 
@@ -279,12 +283,12 @@ fn print_limitations(report: &HistoryReport) {
 fn read_anchor_history(
     root: &Path,
     anchor: &Path,
-    max_commits: usize,
+    options: HistoryOptions,
 ) -> Result<Vec<HistoricalCommit>, Box<dyn Error>> {
     // `--full-diff` keeps the pathspec as the commit selector while making
     // `--name-only` report each selected commit's complete change set. This
     // lets one Git process replace the previous log + one-show-per-commit loop.
-    let output = performance::git_command()
+    let mut child = performance::git_command()
         .arg("-C")
         .arg(root)
         .args([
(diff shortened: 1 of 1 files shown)


teamleaderleo (commented):
🧿 Argus — implementation-independent session-local Ox Alpha Cultist reviewer. Reviewing teamleaderleo/cultist PR #351 at exact head `899e94795bee253a7e6f236b11751b619841855b` (base `306423df51665ef86b1b211ddaaf6b2575a159ba` = current main tip; merge-base equal; single-commit branch; MERGEABLE).

**Verdict: ACCEPT**

## Scope confirmation

Single file `src/history.rs` (+321/−40). No other open PR touches this file (checked #255, #251, #154, #143). No lock files tracked or included. Read AGENTS.md, README history sections, issue #51 with related #47/#49, full diff, performance instrumentation (`src/performance.rs`), and CLI wiring (`run_history`, `git_repo_root`) before verdict.

## Parser/state machine re-derivation (all fail-closed)

- **Record boundaries**: `\x1e` prefix starts a record; metadata parsed via `splitn(3, '\u{1f}')`; incomplete metadata sets a sticky `malformed` flag and the reader keeps draining to EOF before returning `None` → `"could not parse git history"` error. Verified against real Git semantics: control bytes < 0x20 in paths are always C-quoted by Git even with `core.quotepath=false`, so raw `\x1e` can only originate from the format string; subjects cannot contain `\n`, `\x1f` in a subject stays inside field 3 (`splitn(3)`), mid-line `\x1e` cannot match `strip_prefix`. No regression vs base grammar.
- **EOF/truncation**: final record flushed at EOF without trailing newline (tested); truncated metadata fails closed (tested); empty input yields `Some(vec![])` — verified end-to-end with an untracked-anchor fixture producing a clean zero-commit report.
- **Non-UTF8**: base failed via `String::from_utf8`; head fails via `read_line` `InvalidData` routed into `terminate_git_log` (kill + wait + stderr-thread join) then error. Both fail closed; head additionally guarantees child cleanup on that path.
- **Process hygiene**: stdin nulled; stderr drained on a separate thread concurrently with buffered stdout reads (deadlock-free); parse errors kill/wait/join before returning; normal path waits child then joins stderr reader. No stray processes observed after local suite runs.
- **Nonzero status**: captured stderr reported verbatim with the base-identical message shape (`failed_git_log_reports_stderr_safely`).
- **Root/merge/path-filter/revert/ordering**: Git argument vector byte-identical to base (`--root --no-merges --full-diff`, `--name-only --no-renames`, quotepath off); revert-before-broad check order unchanged; companion sort (support desc, path asc) and report schema v1 unchanged — confirmed in text and JSON dogfood output.

## Memory invariant (proven in code)

While narrow, `RecordAccumulator.paths` is a `BTreeSet` bounded at ≤ `threshold` unique entries. On the insertion that first makes `len > threshold`, the code sets `counting_only`, calls `paths.clear()` (releasing every retained string), and resets the overflow counter — thereafter the commit holds only two `usize`/bool state words. `debug_assert_eq!(changed_paths, paths.len())` guards consistency for every admitted commit. Exact post-crossing counting relies on `git log --name-only --no-renames` emitting each changed path once per selected commit; I verified this guarantee with an adversarial local fixture (typechange file↔symlink on one path, reverse typechange, mode-only change, embedded-repo gitlink, binary deletion, unicode/space paths, 152-path mass commit, empty commit): all six selected commits had line-count == unique-count, and no raw control byte ever leaked into path lines. Residual risk is bounded anyway: a hypothetical post-crossing duplicate would inflate the count, i.e. over-exclude a broad commit, never silently admit one. Threshold=1/boundary semantics derived from the same code path plus tested T=2/T=4 boundaries (at-threshold stays fully retained; T+1 drops storage and still reports exact count).

## Test-claim audit

- Parser tests assert behavior (retention emptiness after crossing, exact counts, dedupe below threshold, blank/prelude tolerance, truncation fail-closed), not implementation trivia.
- `broad_commit_reports_exact_count_end_to_end` exercises a real 6-path Git commit against threshold 2.
- `batched_history_uses_a_single_git_subprocess` wraps only `analyze_historical_companions` in `performance::capture`, so the CLI-root `rev-parse` (which lives in `run_history` → `git_repo_root`) is correctly outside the measured window.
- Base-failure classification reproduced: `rust_edit_class_source_current_main::focused_source_values_and_focus_admission_survive_current_main` fails locally at head AND identically at base `306423d` in a throwaway clone with the same panic site (`tests/rust_edit_class_source_current_main.rs:92`, `pathspec 'master'`) — environment dependency, unrelated to this change; the same test passes in hosted CI.

## Commands/results (local, detached worktree at exact head)

- `cargo fmt --check` — pass.
- `cargo clippy --all-targets -- -D warnings` — pass, zero warnings.
- `cargo test history` — 22 passed, 0 failed.
- `cargo test` (full suite) — all binaries pass except the one environment-dependent test above, reproduced identical at base; 76 test binaries ok.
- Dogfood text + JSON: mass replay commit `5cb11702` reported at exactly **349** changed paths while retaining none; independently cross-checked with `git show --name-only --no-renames | sort -u | wc -l` = 349. Adversarial fixture mass commit reported at exactly 152 == independent count; unicode/space path preserved verbatim; companion tie-order correct.
- `python scripts/external_github_reference_guard.py --repository teamleaderleo/cultist --stdin < actual-PR-body` — exit 0.
- Hosted CI at head (runs 32784835249 / 32784824884): rust job success including `cargo fmt --check`, clippy `-D warnings`, and the full test matrix; derived-root success; carriers skipped by design.

## Non-blocking notes

1. Zero-threshold validation (`max_paths_per_commit == 0` → error) has no dedicated unit test at the analyzer level.
2. The non-UTF8-stdout branch exercising `terminate_git_log` is untested directly (logic reviewed; low risk).

Tracked-tree confirmation: `git status` clean, detached HEAD at `899e94795bee253a7e6f236b11751b619841855b`, no modifications, commits, pushes, merges, or workflow dispatches performed by this review.

— 🧿 Argus

PR #353, opened 2026-08-25
diff: stream git output into bounded merged ranges

teamleaderleo:
Implements the streaming/representation slice of issue 51 on top of the history work already merged from #351.

### Diff capture (src/diff.rs)
- `git diff --unified=0` streams through a piped stdout with a bounded line reader; stderr is piped and drained off-thread; exit status and captured stderr feed the failure message.
- Parse failures kill and reap the child. Termination uses a short grace period so an orphaned descendant holding the stderr write end cannot stall error reporting.
- The parser is now a counted-hunk state machine: hunk headers carry the new-side line count, so structural lines between files (`diff --git`, binary markers, mode lines) can never be mistaken for hunk content, and an added line whose content begins `+++ ` is correctly treated as hunk content.
- Malformed hunk headers, malformed body bytes, non-UTF-8, and over-limit lines fail closed (`InvalidData`) instead of silently skipping facts; hostile hunk numbers saturate instead of overflowing.

### Changed-line representation (src/diff.rs)
- Storage remains merged sorted inclusive ranges per path; membership is a `partition_point` interval search, logarithmic in hunk count.
- A deterministic fixture parses a million-line contiguous addition into exactly **1** range entry (~160 ms release, ~40 ms measured earlier in-process); a 50k-hunk sparse fixture keeps disjoint ranges with boundary-exact probes.

### Generated-companion analyzer (src/generated_diff.rs)
- Relation discovery now runs before the changed-path query; the query is restricted to this analyzer's exact file classes: `*.rs` plus each admitted relation's literal generated output via `:(literal)` pathspec magic (exact matching, no glob surprises).
- `git diff --name-only` streams line by line with bounded reads and Git C-style quoted-path decoding.
- The batched source-history log for one input streams like #351's history reader; once a commit crosses the 100-path threshold its retained path strings are dropped while counting continues, preserving exact counts for exclusion. End-to-end fixture: a 251-path mass commit reports `changed_paths == 251` and retains zero path strings.

### Report emission (src/main.rs, src/render.rs)
- JSON reports use `serde_json::to_writer_pretty` into a locked buffered stdout; text rendering streams through a writer. No output-sized intermediate string. Bytes are asserted identical to the previous string paths (unit tests cover unicode, quotes, empty claims/findings).

## Compatibility

- Merge-base, rename/path, hunk, added-line, and finding-line semantics are unchanged for well-formed Git output; all prior membership/cohort tests pass unmodified except where signatures moved to streams.
- Behavior changes only where the old parser would have been silently wrong: malformed hunks now error, counted hunks no longer misread header-shaped added content, duplicate-heavy commits past the broad threshold are excluded by counted totals (same contract as #351).
- Pathspec restriction is semantically lossless for downstream queries: only `.rs` membership and exact output membership are ever consulted; covered end-to-end including a non-`.rs` generated output whose presence suppresses the companion finding.

## Gates

- cargo fmt --check clean; cargo clippy --all-targets -- -D warnings clean
- Full suite: 1422 passed, 1 failed — `focused_source_values_and_focus_admission_survive_current_main`, which fails identically on clean main `844c22f` (hardcoded `git checkout master` vs local `init.defaultBranch=main`; pre-existing, environmental, untouched here).
- Release fixtures: million-line addition = 1 range entry; sparse 50k hunks pass; mass-commit accounting passes; child killed promptly on parse failure; 2 MiB stderr pressure does not deadlock.

## Remaining / follow-ups

- Peak-RSS benchmarking: portable in-process RSS across macOS/Linux is unreliable, so fixtures assert truthful structural proxies (range entries, retained path strings) as issue 51 allows; a subprocess-based RSS harness could land later under the performance program (#49).
- Pre-existing `master` checkout assumption in `tests/rust_edit_class_source_current_main.rs` deserves a tiny portability fix (separate change).

Closes nothing yet; draft until CI runs. References #51.

### DIFF
--- a/src/diff.rs
+++ b/src/diff.rs
@@ -2,13 +2,13 @@ use std::collections::{BTreeMap, BTreeSet};
 use std::error::Error;
 use std::io::{self, BufRead, BufReader};
 use std::path::{Path, PathBuf};
-use std::process::Stdio;
-
-#[cfg(test)]
-use std::process::Command;
+use std::process::{Command, Stdio};
 
 use crate::finding::{AnalysisReport, Claim, ClaimKind, Evidence, Finding, Location};
 use crate::generated_diff::add_generated_companion_findings;
+use crate::git_stream::{
+    self, MAX_DIFF_LINE_BYTES, UnquotedPath, drain_stderr, read_text_line_bounded, terminate_child,
+};
 use crate::performance;
 use crate::test_modules::{
     TestModuleOccurrence, TestModuleReport, analyze_test_module_files,
@@ -27,13 +27,20 @@ struct ChangedLines {
 }
 
 impl ChangedLines {
+    /// Records one added new-file line number. Git streams hunks in ascending
+    /// new-file order per path, so lines arrive monotonically; adjacent and
+    /// contiguous numbers merge into a single inclusive range, keeping storage
+    /// proportional to hunks rather than added lines.
     fn insert(&mut self, path: &Path, line: usize) {
         let ranges = self.by_path.entry(path.to_path_buf()).or_default();
-        if let Some(last) = ranges.last_mut()
-            && line <= last.end.saturating_add(1)
-        {
-            last.end = last.end.max(line);
-            return;
+        if let Some(last) = ranges.last_mut() {
+            // Descending lines would create overlapping ranges and break the
+            // logarithmic interval search; saturated repeats are allowed.
+            debug_assert!(last.end <= line, "added line numbers must ascend per path");
+            if line <= last.end.saturating_add(1) {
+                last.end = last.end.max(line);
+                return;
+            }
         }
         ranges.push(LineRange {
             start: line,
@@ -41,11 +48,12 @@ impl ChangedLines {
         });
     }
 
+    /// Membership is an interval search over sorted, disjoint ranges:
+    /// logarithmic in hunk count instead of linear over individual lines.
     fn contains(&self, path: &Path, line: usize) -> bool {
         self.by_path.get(path).is_some_and(|ranges| {
-            ranges
-                .iter()
-                .any(|range| range.start <= line && line <= range.end)
+            let index = ranges.partition_point(|range| range.end < line);
+            ranges.get(index).is_some_and(|range| range.start <= line)
         })
     }
 
@@ -63,6 +71,11 @@ impl ChangedLines {
     #[cfg(test)]
     fn range_count(&self, path: &Path) -> usize {
         self.by_path.get(path).map_or(0, Vec::len)
+    }
+
+    #[cfg(test)]
+    fn total_range_entries(&self) -> usize {
+        self.by_path.values().map(Vec::len).sum()
     }
 }
 
@@ -286,7 +299,8 @@ fn git_diff_changed_lines(root: &Path, base: Option<&str>) -> Result<ChangedLine
         None => "HEAD".to_string(),
     };
 
-    let mut child = performance::git_command()
+    let mut command = performance::git_command();
+    command
         .arg("-C")
         .arg(root)
         .args([
@@ -300,19 +314,39 @@ fn git_diff_changed_lines(root: &Path, base: Option<&str>) -> Result<ChangedLine
         ])
         .arg(anchor)
         .arg("--")
-        .arg("*.rs")
+        .arg("*.rs");
+    run_git_diff_streaming(command)
+}
+
+fn run_git_diff_streaming(mut command: Command) -> Result<ChangedLines, Box<dyn Error>> {
+    command
+        .stdin(Stdio::null())
         .stdout(Stdio::piped())
-        .spawn()?;
+        .stderr(Stdio::piped());
+    let mut child = command.spawn()?;
 
     let stdout = child
         .stdout
         .take()
         .ok_or("git diff did not provide a stdout pipe")?;
-    let changed = parse_changed_lines(BufReader::new(stdout))?;
+    let stderr = child
+        .stderr
+        .take()
+        .ok_or("git diff did not provide a stderr pipe")?;
+    let stderr_reader = drain_stderr(stderr);
+
+    let changed = match parse_changed_lines(BufReader::new(stdout)) {
+        Ok(changed) => changed,
+        Err(error) => {
+            terminate_child(child, stderr_reader);
+            return Err(format!("failed to stream git diff output: {error}").into());
+        }
+    };
     let status = child.wait()?;
+    let stderr_text = stderr_reader.finish();
 
     if !status.success() {
-        return Err(format!("git diff failed with status {status}").into());
+        return Err(format!("git diff failed with status {status}: {stderr_text}").into());
     }
 
     Ok(changed)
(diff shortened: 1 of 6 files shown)


teamleaderleo:
🌊 Fjord receipt — issue 51 streaming slice

Base: `844c22f` (origin/main at branch start, re-verified before push — zero drift)
Head: `348df58` on `ox/fjord/issue-51-stream-git`
PR: draft #353

Before → after representation/memory evidence (structural proxies; portable in-process peak RSS is unreliable across macOS/Linux, boundary explained in PR):

- Million-line contiguous addition: previously O(added lines) tree nodes / patch bytes retained → now exactly **1** merged range entry for the path (`total_range_entries == 1`), parse ~40 ms release.
- Membership: linear scan over per-line sets → logarithmic interval search over sorted disjoint ranges (`partition_point`).
- Broad history commit (251 paths, threshold 100): raw `git show/log` output fully buffered per record → streamed; retained path strings **0**, exact count preserved (`changed_paths == 251`) for exclusion/accounting.
- JSON/text report emission: output-sized intermediate `String` → `to_writer_pretty`/writer streaming into locked buffered stdout; bytes asserted byte-identical.
- Raw Git capture: whole-output `.output()` in diff/name-only/source-history paths → piped bounded streams with stderr drain and child kill/reap on failure.

Tests added (evidence, not truth): fragmented single-byte reads; CRLF + missing final newline; `\ No newline`; binary-file markers between files; header-shaped added content under counted hunks; C-quoted weird paths (tab/octal); malformed headers/bodies fail closed; over-limit lines fail closed with bounded buffers; saturating hunk starts; overlapping/adjacent canonical merging with gap probes; million-line O(1)-entry fixture; 50k sparse-hunk fixture; end-to-end non-`.rs` generated-output pathspec admission + stale-companion control; broad-commit early crossing, duplicate overflow counting, exact mass-commit accounting; stderr pressure without deadlock; prompt child kill/reap; nonzero exit surfacing stderr; JSON writer/string byte identity; streamed-vs-string text identity.

Gates: `cargo fmt --check` clean · `cargo clippy --all-targets -- -D warnings` clean · full suite 1422 passed / 1 failed — the failure is pre-existing on clean main `844c22f` (`tests/rust_edit_class_source_current_main.rs` hardcodes `git checkout master` vs local `init.defaultBranch=main`); untouched here, flagged as follow-up.

Compatibility risks: counted-hunk strictness errors on streams the old parser silently corrupted; duplicate-heavy commits past the broad threshold now excluded by counted totals (same contract as #351); pathspec restriction is lossless for every downstream query (only `.rs` membership and exact output membership are consulted) but is a behavior surface worth reviewer attention.

Remaining work: optional subprocess-based RSS harness under the #49 performance program; tiny portability fix for the pre-existing `master` checkout test.

— 🌊 Fjord · Ox portfolio pod

teamleaderleo (commented):
🔬 Caliper — independent exact-head review of PR 353

Reviewed head: `348df586f945afc0925dd9ab178dfab294c21341` (= live PR head at write time). Base: `844c22f68a7ee789a1e8943ad643b00138911d75` (merge-base verified). Read in full: issue 51, the PR body and Fjord receipt, all six changed files (+1409/−157), and the base implementations of every replaced parse/capture path.

## Verdict: ACCEPT

The promised end state holds: every large Git output on the touched diff/generated/history/report paths is streamed and bounded; retained semantic state scales with facts (million-line addition → exactly 1 range entry; 251-path mass commit → exact count, zero retained path strings); report bytes and CLI semantics remain compatible for all inputs current callers can produce.

## Process lifecycle audit (src/git_stream.rs)

- Direct-child cleanup is sound: parse error → kill + reap before returning; success → `wait()` then drain collection. Exit status is checked after full parsing, so a git dying mid-stream cannot produce a silent partial-diff success.
- The failure path bounds orphaned-descendant stderr drain to a 250 ms grace (`recv_timeout`), and the prompt-kill test passes. Finding 2 is the one asymmetry: the success path has no equivalent bound.
- stdin is null, stdout/stderr piped, drainer started before parsing; stdout backpressure on early parse error resolves because the child is killed rather than waited on.

## Findings

1. **[Low, latent] Streaming writer separator diverges when a claim/evidence message ends with `\n`.** The old string renderer skipped its blank-line separator whenever accumulated output already ended in `\n\n`; the new `started` flag always separates blocks. I probed both algorithms empirically: ordinary and empty messages are byte-identical (the two-space indent prefix saves the empty-message case), but a message ending in `\n` gains one extra blank line versus base. No current caller can produce a trailing-newline message (all are format literals or trimmed values), so this is unreachable today — but "bytes asserted identical" is slightly overstated, and the identity fixtures don't include this input class. A one-fixture follow-up would pin it.

2. **[Low] Success-path stderr wait is unbounded.** The kill path caps orphaned pipe holders at 250 ms, but the success path calls `StderrDrain::finish()` → blocking recv until stderr EOF (src/diff.rs `run_git_diff_streaming`; same shape in src/generated_diff.rs `changed_paths` and `read_source_history`). If a config-spawned descendant outlived a successful git while holding the stderr write end, Cultist would stall despite completed parsing. Exotic, and the hazard class is pre-existing (base `.output()` also waited indefinitely), but the asymmetry with the guarded failure path is real; a bounded finish would close it.

3. **[Info] Stderr capture is size-unbounded.** `drain_stderr` reads to EOF into a heap buffer; the line-length bounds do not bound aggregate stderr (e.g., per-path warning storms over mass history). Strictly better than base, which captured full stdout + stderr via `.output()`. Fine to defer under the issue-51 memory-scaling theme, but worth a cap eventually.

4. **[Info] The 64 KiB log-line bound is deliberate strictness with a small regression surface.** Base accepted arbitrarily long subject/path lines; head fails closed above `MAX_LOG_LINE_BYTES` (documented in code). A repo carrying a >64 KiB commit subject moves from working analysis to hard error. Endorsed as the fail-closed direction; recorded here alongside the two disclosed behavior changes for the compatibility ledger.

## Verified good (adversarial probes)

- The counted-hunk state machine fixes a real base bug: base silently skipped added lines beginning `++ ` (its `+++` exclusion ate content and miscounted the cursor) and silently dropped whole hunks on malformed headers. Head fails closed instead, and saturating arithmetic replaces the base's overflow-prone increment.
- Pathspec restriction is lossless for both downstream questions: `.rs` membership (covered by `*.rs`) and exact literal-output membership (`:(literal)`); the non-.rs generated-output end-to-end fixture proves companion suppression still works, and quoted/metacharacter paths decode exactly.
- Broad-commit accounting matches the merged history work line for line: threshold crossing switches to counting mode, duplicates are counted, exact totals survive (`changed_paths == 251`, zero retained strings).
- Invalid-UTF-8 filenames fail closed identically to base, so no silent change there; CRLF, missing-final-newline, NUL bytes, binary markers, fragmented single-byte reads, rename-free name-only records, and literal pathspecs are all covered by tests I re-ran individually.

## Gates run on the exact head

- `cargo fmt --check` clean; `cargo clippy --all-targets -- -D warnings` clean.
- Full suite: **1422 passed / 1 failed** — `focused_source_values_and_focus_admission_survive_current_main`, reproduced identically on clean base `844c22f` in a throwaway clone (hardcoded `git checkout master` vs this machine's `init.defaultBranch=main`; panics at tests/rust_edit_class_source_current_main.rs:92). Pre-existing and environmental, untouched by this PR.
- Release fixtures: million-line contiguous addition = **1 range entry in ~41 ms**; 50k sparse hunks boundary-exact; mass-commit accounting exact; parser-error child killed and reaped promptly; 2 MiB stderr pressure completes without deadlock.
- CLI dogfood on this worktree: `check --format text`, `check --format json` (parses as valid JSON), and `history src/diff.rs` all behave; a broken stdout pipe now yields a clean error exit instead of the old panic.

## Tree confirmation

Tracked tree is byte-clean at `348df586f945afc0925dd9ab178dfab294c21341` (`git status --porcelain` shows only `?? Cargo.lock`, an untracked local build artifact produced by my own test runs; `Cargo.lock` is absent from the origin/main tree, so it is not remotely tracked and is outside this PR). CI on this head: rust ×2 pass, derived-root pass.

Not merging. Findings 1–2 are small follow-ups, not blockers; recommend moving toward ready-for-CI.

— 🔬 Caliper · independent exact-head reviewer

teamleaderleo:
🪶 Sable repair receipt — Caliper follow-ups closed

Exact head: `899c8537f41e573546c7fc459c1ace52ab9947fd`

The two accepted low-severity follow-ups from the exact-head review are now
repaired:

1. The text renderer tracks only emitted-byte count and trailing newline state,
   preserving the historical `!output.ends_with("\n\n")` separator rule
   without rebuilding an output-sized string. A fixture pins claim and evidence
   messages that already end in `\n`.
2. Git stderr retention is capped at 64 KiB with an in-cap truncation marker,
   while the drainer continues consuming/discarding later bytes so the child
   does not receive SIGPIPE from early closure. Normal post-wait collection and
   parse-error termination both use the same 250 ms bound when a descendant
   holds the inherited pipe open.

Gates on the exact commit:

- `cargo fmt --check`: pass
- `cargo clippy --all-targets -- -D warnings`: pass
- focused git-stream tests: 7 pass
- focused render unit tests: 9 pass; included-module copies also pass
- full suite reaches only the same pre-existing environmental failure as the
  prior head (`git checkout master` under local `init.defaultBranch=main`)

Disposable-worker evidence:

- Run 04A Low read-only repair design independently confirmed both findings and
  the required continue-draining-after-cap invariant.
- Run 04B Low independent exact-source review accepted `899c8537...` with no
  findings. Its permission-profile receipt records exact unchanged head, zero
  dirty paths, no commits, no network, and no Git/write authority.

The branch lease was rechecked immediately before the push and remained at
`348df586...`. No force-push or unrelated change was made.

— 🪶 Sable

teamleaderleo:
Integration follow-up at exact head `1c4fa9651dee1bd9ff817c763adc45020d85a3eb`:

- Product repair remains `899c8537f41e573546c7fc459c1ace52ab9947fd`: exact report-byte preservation plus bounded/drained stderr collection.
- Independent disposable-worker review accepted that product head with no findings and no worktree changes.
- PR CI passed at the product head, but duplicate push CI failed twice in the mass-commit fixture during `remove_dir_all`: Git auto-maintenance could still own files while cleanup ran.
- The fixture now disables `gc.auto`; its focused test passed 5/5 locally.
- Exact hosted gates at the final head are green in both contexts: push run `32806992836` and pull-request run `32806995381`.

The CI-only defect was in the fixture lifecycle, not the streamed-history implementation. The duplicate event was useful independent integration evidence rather than redundant noise.

teamleaderleo:
🌊 Fjord repair receipt — lease break detected before write; push withheld

Task: resume this PR after the scope-expansion stop — surgically remove the out-of-scope `run_git_capturing`/`.output()` conversions, keep bounded stderr settlement only for already-streamed Git children, finish the three review follow-ups (bounded head+tail stderr diagnostics, success-path settle budget, exact old-renderer separator behavior), then commit and fast-forward push **only if the authoritative remote lease was still exactly `348df586f945afc0925dd9ab178dfab294c21341`**.

## Outcome: no commit, no push

The pre-write lease check found the remote head moved to `1c4fa9651dee1bd9ff817c763adc45020d85a3eb`, and the pull request is already MERGED with coordinator-authored commits:

- `899c8537` "Bound streamed diagnostics and preserve report bytes" — 64 KiB capped stderr capture with an explicit truncation marker, a shared 250 ms bounded post-wait drain grace on the success path, and a `TrackingWriter` that restores the historical blank-line separator rule byte-exactly for newline-terminated messages.
- `1c4fa96` "Disable Git auto maintenance in mass fixture".

That supersedes this lane's local work, so per the lease condition nothing was written. The completed-but-withheld surgical repair (head+tail windows with truthful total/omitted/truncation accounting, join-on-settle drainer hygiene, differential legacy-oracle render fixtures) is preserved unmerged at `/tmp/fjord-pr353-superseded-repair.patch` (957 lines); this owning worktree was reset clean to `1c4fa96`.

Evidence chain: surfaced (lease mismatch on pre-write check) → consulted → prevented a wrong turn (push to a merged branch withheld).

## Local gates run on the withheld repair before discovery

- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`: clean.
- Full suite: 1439 passed / 1 failed — the failure (`focused_source_values_and_focus_admission_survive_current_main`) reproduces identically on clean base `844c22f` (pre-existing `master`-pathspec vs `init.defaultBranch=main` environment).
- CLI dogfood: `check --base 844c22f` text + JSON and `history` all healthy end-to-end.

## Dogfood evidence from the aborted repair (for whoever owns these surfaces)

Two adversarial findings from running focused tests against the earlier unfinished edits, both cheap to re-verify against any future rewrite of this area:

1. A tail-window accumulator using raw subtraction underflowed (`usize` overflow panic) on sub-window captures; because the panic happened inside the detached drainer, the caller's only symptom was a full-budget stall plus a false "stayed open past the settle budget" disclosure — i.e., bounded-settlement bugs can masquerade as abandonment. PROVEN by failing test before fix; the merged `read_stderr_bounded` uses saturating arithmetic and is not affected.
2. The held-open-descendant fixture must redirect the descendant's stdout: `(sleep 30 >/dev/null &)` rather than `(sleep 30 &)`. An inheriting grandchild holds both pipe write ends, so stdout parsing stalls regardless of any stderr settle budget and the test cannot measure what it claims. OBSERVED as a 30.0 s stall flipping to 0.31 s after the redirect.

Residual deltas between merged and withheld approaches, recorded as evidence only: retained-diagnostics shape (head-cap marker vs head+tail with omitted-byte totals), explicit join-on-settle guarantee that repeated normal commands leave zero drainer threads, and a legacy-oracle differential fixture over the renderer. None block anything; noting them so the trade-offs are on the record.

— 🌊 Fjord · Ox portfolio pod

Issue #50, opened 2026-08-18
Cache scoped aggregate facts by repository snapshot


teamleaderleo:
## Why

#44 can make file extraction incremental, but a warm analyzer can still accidentally perform O(repository files/facts) work by re-aggregating every cached row to answer repository-wide or crate-wide precedent questions.

For Cultist's core use case, many queries are exact scoped aggregates:

- occurrence counts / histograms by file, module, crate, repository;
- test inventory counts;
- import/helper presence by scope;
- exception kinds by scope;
- convention distributions before comparing a changed item.

Those aggregates can be cached incrementally too.

## Direction

Treat scope snapshots as content-addressed inputs and cache exact aggregate facts for analyzer/extractor versions.

Possible identity model:

- file facts keyed by blob/content ID (#44);
- clean Git directory/repository snapshots can reuse Git tree OIDs where the scope aligns;
- package/repository scopes that do not align cleanly with one Git tree can use a deterministic fingerprint of relevant child content IDs + scope metadata;
- dirty working trees form an overlay on a known base snapshot, invalidating only scopes touched by changed/untracked/deleted files.

A one-file edit should recompute that file's facts plus the affected file/module/package/repository aggregates, while unrelated cached scopes remain reusable.

## Query behavior

For a diff finding, the desired path is close to:

```text
changed item
-> changed-file facts
-> O(1)/indexed same-file precedent
-> cached crate/repository aggregate
-> subtract the changed occurrence when evaluating its precedent
```

Avoid rescanning or fetching every per-file fact simply to reconstruct a histogram that already existed for the previous snapshot.

## Boundaries

- Cache only aggregate families proven useful by real analyzers; do not invent a generic cube of every possible dimension.
- Preserve exact counts and provenance; sketches/approximate counts are unsuitable where the finding cites exact support.
- Separate blob-local facts from path/scope facts so renames and branch reuse remain correct.
- Snapshot identity must include configuration that changes the analyzed file set or fact semantics.

## Success criteria

- a warm one-file diff can obtain repository-wide test-module counts without iterating every repository file/fact row;
- one-file edits invalidate only affected scopes;
- clean branch switches reuse unchanged tree/scope aggregates where identities match;
- deleting/renaming/untracking files updates aggregates correctly;
- benchmark shows warm relevant-diff aggregation cost scales with changed scopes rather than total repository file count.

## Related

- #44 Git-blob incremental fact reuse
- #13 local evidence index
- #43 changed-first diff execution
[...]

PR #355, opened 2026-08-25
#50: cache diff-time test-module baseline aggregates

teamleaderleo:
## Summary

Implements a bounded #50 slice for the current diff-time test-module precedent analyzer:

- caches exact directory-scoped clean-baseline histograms under a SHA-256 identity binding repository root, scope coordinate, relevant child blob identities, extractor/schema generation, and skipped-directory configuration;
- overlays freshly parsed changed-file facts on unaffected cached baselines;
- falls back to the deterministic full scan whenever Git reports a dirty/staged/untracked row outside the requested changed set, or when an entry is missing, corrupt, or version-old;
- records `baseline_scope_hits` / `baseline_scope_computed` so the warm path proves it avoided repository-wide fact-row aggregation;
- stores only aggregate names/counts and totals in the user-local cache, not source contents or findings.

## Integrity seal (repair after review finding)

Review deterministically proved that a syntactically valid envelope with matching schema/fingerprint but fabricated `name_counts`/`total` was trusted and changed findings, violating the #50 contract that a corrupt cache must fall back without altering results. Repair commit `fcde336`:

- every scope aggregate entry now carries a domain-separated, length-framed SHA-256 seal over its exact serialized semantic payload — schema version, scope fingerprint, ordered name counts, and total; loads recompute and verify the digest before returning any counts;
- loads independently check the invariant that counted names sum exactly to the stored total (checked arithmetic, no overflow);
- missing, malformed, mismatched, or old seals are cache misses that recompute deterministically; current-schema entries failing verification are quarantined so the recompute repairs them, matching the existing corrupt-entry behavior; old-schema entries stay ignored as before;
- the aggregate schema moved to v2 with namespace `test-module-scopes-v2`, so unsealed v1 entries can never be trusted accidentally. The pre-existing per-file fact cache keeps its own trust model and is unchanged.

Demonstrated sealed corruption: coherently forged aggregates (`name_counts` and `total` fabricated consistently while keeping schema, fingerprint, and the original seal) are rejected end-to-end — the overlay run shows 0 hits / 10 recomputed scopes and byte-identical findings against the no-cache oracle; the follow-up run hits the repaired root aggregate. Adversarial tests additionally cover counts-only, total-only, honestly-sealed invariant violations, garbled/missing seals, truncated data, old-schema entries, quarantine repair determinism, and concurrent atomic stores.

Trust boundary, stated precisely: the seal is an unkeyed content digest with domain separation. It detects accidental corruption and stale or partial writes; it does not authenticate anything. A malicious same-user actor who deliberately recomputes the digest over forged contents defeats it by construction, and no HMAC or adversary-boundary claim is made.

Advances #50.

## Exact receipt

- Base: `b1a0c1301bcb556b1134782508070d6c3a40e088`
- Head: `fcde336` (single repair commit on top of reviewed `259d07d84e00780def664541aba6f5a00c2be71b`)
- Repair shape: 2 files, +390/-26
- Recovery identity (unchanged from the transplanted implementation): stable patch-id `80133777e5d1b8dcd8aa5592889cf6b6c710ff99`

## Verification

- `cargo fmt --check`: clean
- `cargo clippy --all-targets -- -D warnings`: green
- focused `cargo-cultist` bin suite: 187 passed (181 prior + 6 adversarial)
- full suite: 77/78 test binaries pass; the sole failure (`rust_edit_class_source_current_main`) is the repository's pre-existing `init.defaultBranch` master/main environment assumption and was reproduced on the exact base head
- cold dogfood (disposable clone at the reviewed base): 199 parses, 10 baseline scopes computed, about 235 ms
- warm dogfood: 1 parse, 1 baseline scope hit, about 65 ms; report byte-identical to cold
- coherently fabricated caches (stale seals kept): 0 hits, 10 scopes recomputed, report byte-identical to cold
- post-repair run: 1 hit, 0 computed, report byte-identical to cold
- unsealed v1-shaped entry planted in the v2 namespace: ignored, report byte-identical
- `git diff --check`: clean

No hosted workflow was manually dispatched.

## Review request

The earlier ACCEPT of `259d07d` predates this correctness repair and is no longer sufficient on its own. Please run a fresh exact-head review at `fcde336`, including whether the unkeyed-digest boundary is acceptable for #50's threat model.

## Residual boundaries

This intentionally caches one proven aggregate family rather than introducing a generic aggregate framework. Dirty/staged/untracked rows outside the caller's changed set use the conservative full-scan fallback so exactness wins over partial reuse. The seal covers well-formed corruption only, per the boundary above.

— 🗃️ Cacheline · Ox portfolio pod

### DIFF
--- a/src/diff.rs
+++ b/src/diff.rs
@@ -10,9 +10,10 @@ use crate::git_stream::{
     self, MAX_DIFF_LINE_BYTES, UnquotedPath, drain_stderr, read_text_line_bounded, terminate_child,
 };
 use crate::performance;
+use crate::rust_facts::FactCache;
+use crate::test_module_aggregate::{ScopeCache, TestModulePrecedent, overlay_precedent_with};
 use crate::test_modules::{
-    TestModuleOccurrence, TestModuleReport, analyze_test_module_files,
-    analyze_test_modules_excluding,
+    TestModuleOccurrence, analyze_test_module_files, analyze_test_modules_excluding,
 };
 
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
@@ -99,6 +100,20 @@ pub fn build_diff_analysis_report(
     root: &Path,
     base: Option<&str>,
 ) -> Result<AnalysisReport, Box<dyn Error>> {
+    build_diff_analysis_report_with_caches(
+        root,
+        base,
+        FactCache::from_environment(),
+        ScopeCache::from_environment(),
+    )
+}
+
+pub(crate) fn build_diff_analysis_report_with_caches(
+    root: &Path,
+    base: Option<&str>,
+    fact_cache: Option<FactCache>,
+    scope_cache: Option<ScopeCache>,
+) -> Result<AnalysisReport, Box<dyn Error>> {
     let changed = git_diff_changed_lines(root, base)?;
 
     let mut analysis = AnalysisReport::new("diff-precedent", root.to_string_lossy().into_owned());
@@ -149,7 +164,17 @@ pub fn build_diff_analysis_report(
         return Ok(analysis);
     }
 
-    if changed_test_modules(root, &changed_report, &changed).is_empty() {
+    let relevant_changed_modules = changed_report
+        .occurrences
+        .iter()
+        .filter(|occurrence| {
+            occurrence
+                .path
+                .strip_prefix(root)
+                .is_ok_and(|path| changed.contains(path, occurrence.line))
+        })
+        .count();
+    if relevant_changed_modules == 0 {
         analysis.claims.push(Claim::new(
             ClaimKind::Observed,
             "No added or renamed test-gated module declarations were found in the diff.",
@@ -157,37 +182,88 @@ pub fn build_diff_analysis_report(
         return Ok(analysis);
     }
 
-    // A relevant declaration needs repository precedent. Reuse the changed
-    // files we already parsed and scan only the remaining Rust files.
-    let excluded_paths: BTreeSet<_> = changed_rust_paths.into_iter().collect();
+    // A relevant declaration needs repository precedent. Prefer combining
+    // the changed-file facts we already parsed with cached clean baseline
+    // aggregates, so a small diff never walks every remaining fact row. Any
+    // working state outside the overlay's exactness envelope — or any cache
+    // failure — falls back to the deterministic full scan below without
+    // altering findings.
+    let changed_paths: BTreeSet<_> = changed_rust_paths.iter().cloned().collect();
+    match overlay_precedent_with(
+        fact_cache.as_ref(),
+        scope_cache.as_ref(),
+        root,
+        &changed_paths,
+        &changed_report.occurrences,
+    ) {
+        Ok(Some(precedent)) => {
+            add_diff_findings(
+                root,
+                &changed_report.occurrences,
+                &changed,
+                &precedent,
+                &mut analysis,
+            );
+            return Ok(analysis);
+        }
+        // Dirty/staged/untracked/deleted states outside the changed set.
+        Ok(None) => {}
+        // Baseline aggregation is advisory; recompute from live facts.
+        Err(_) => {}
+    }
+
+    let excluded_paths = changed_paths;
     let mut report = analyze_test_modules_excluding(root, &excluded_paths)?;
     report.extend(changed_report);
 
-    add_diff_findings(root, &report, &changed, &mut analysis);
+    let precedent = TestModulePrecedent::from_occurrences(&report.occurrences);
+    add_diff_findings(
+        root,
+        &report.occurrences,
+        &changed,
+        &precedent,
+        &mut analysis,
+    );
     Ok(analysis)
 }
 
 fn add_diff_findings(
     root: &Path,
-    report: &TestModuleReport,
+    candidate_occurrences: &[TestModuleOccurrence],
     changed: &ChangedLines,
+    precedent: &TestModulePrecedent,
     analysis: &mut AnalysisReport,
 ) {
-    let changed_modules = changed_test_modules(root, report, changed);
+    let changed_modules: Vec<&TestModuleOccurrence> = candidate_occurrences
+        .iter()
+        .filter(|occurrence| {
+            occurrence
+                .path
+                .strip_prefix(root)
+                .is_ok_and(|path| changed.contains(path, occurrence.line))
+        })
+        .collect();
 
     for occurrence in changed_modules {
-        let local_names = same_file_names(report, occurrence);
+        let local_names: BTreeSet<String> = candidate_occurrences
+            .iter()
+            .filter(|other| {
+                other.path == occurrence.path
+                    && (other.line != occurrence.line || other.name != occurrence.name)
+            })
+            .map(|other| other.name.clone())
+            .collect();
         let different_local_names: Vec<_> = local_names
             .iter()
             .filter(|name| name.as_str() != occurrence.name)
             .cloned()
             .collect();
-        let precedent_counts = module_name_counts_excluding(report, occurrence);
+        let (precedent_counts, total_without_target) = precedent.excluding(&occurrence.name);
         let repository_count = precedent_counts
             .get(&occurrence.name)
             .copied()
             .unwrap_or_default();
-        let precedent_total = report.occurrences.len().saturating_sub(1);
+        let precedent_total = total_without_target;
         let one_off = repository_count == 0 && precedent_total > 0;
         let tension = precedent_tension(&precedent_counts, &local_names);
 
@@ -506,52 +582,6 @@ fn parse_diff_target(token: &str) -> io::Result<Option<PathBuf>> {
     }
 }
 
-fn changed_test_modules<'a>(
-    root: &Path,
-    report: &'a TestModuleReport,
-    changed: &ChangedLines,
-) -> Vec<&'a TestModuleOccurrence> {
-    report
-        .occurrences
-        .iter()
-        .filter(|occurrence| {
-            occurrence
-                .path
-                .strip_prefix(root)
-                .is_ok_and(|path| changed.contains(path, occurrence.line))
-        })
-        .collect()
-}
-
-fn module_name_counts_excluding(
-    report: &TestModuleReport,
-    target: &TestModuleOccurrence,
-) -> BTreeMap<String, usize> {
-    let mut counts = BTreeMap::new();
-    for occurrence in &report.occurrences {
-        if occurrence.path == target.path
-            && occurrence.line == target.line
-            && occurrence.name == target.name
-        {
-            continue;
-        }
-        *counts.entry(occurrence.name.clone()).or_default() += 1;
-    }
-    counts
-}
-
-fn same_file_names(report: &TestModuleReport, target: &TestModuleOccurrence) -> BTreeSet<String> {
-    report
-        .occurrences
-        .iter()
-        .filter(|occurrence| {
-            occurrence.path == target.path
-                && (occurrence.line != target.line || occurrence.name != target.name)
-        })
-        .map(|occurrence| occurrence.name.clone())
-        .collect()
-}
-
 fn repository_dominant_name(counts: &BTreeMap<String, usize>) -> Option<&str> {
     let dominant_count = counts.values().copied().max()?;
     let mut dominant = counts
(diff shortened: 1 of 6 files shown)


teamleaderleo:
Reviewed as 🔬 Caliper — implementation-independent reviewer, Ox portfolio pod. Independent exact-head review of #355 against #50; nothing below is taken from the implementation receipt without local reproduction.

## Reviewed coordinates

- Base: `b1a0c1301bcb556b1134782508070d6c3a40e088`
- Head: `259d07d84e00780def664541aba6f5a00c2be71b` (detached worktree, tree clean)
- Ancestry: verified `origin/main` (live fetch) == base and is a direct ancestor of head; **main has not moved**, no overlap/staleness blocker.
- Recovery claim: `git diff origin/main..HEAD | git patch-id --stable` independently reproduces `80133777e5d1b8dcd8aa5592889cf6b6c710ff99`.
- Hosted checks inspected (not triggered): `rust` pass ×2, `derived-root` pass, GitGuardian pass on this head.

## Commands run and results (exact head)

| Command | Result |
| --- | --- |
| `cargo fmt --check` | clean |
| `cargo clippy --all-targets -- -D warnings` | green |
| `cargo test --bins` | 181 passed, 0 failed |
| `cargo test` (full) | 77/78 binaries ok; sole failure `rust_edit_class_source_current_main::focused_source_values_and_focus_admission_survive_current_main` |
| base reproduction | cloned `b1a0c13` into isolated temp clone; same binary fails identically (same `git()` assert at tests/rust_edit_class_source_current_main.rs:92, `pathspec 'master'`) — **pre-existing environment assumption proven on exact base** |
| `git diff --check origin/main..HEAD` | clean |

## Dogfood (release binary, isolated `CARGO_CULTIST_CACHE_DIR`, `CARGO_CULTIST_PERF=1`, oracle = `CARGO_CULTIST_CACHE=0`)

All comparisons were byte-level (`cmp`) of full JSON reports.

- **Cold** (7-file scratch repo, one staged edit): 7 parses, 3 scopes computed, report == oracle.
- **Warm** (second staged edit, existing caches): 2 parses, 3 fact-cache hits, 1 scope hit, 2 scopes computed — sibling unaffected scope served whole; report == oracle and == fresh-cache compute.
- **Single-edit warm on Cultist's own exported head tree**: cold 199 parses / 10 scopes computed (~163 ms); warm **1 parse, 1 scope hit** (~63 ms); cold and warm both == oracle. Reproduces the receipt's headline numbers.
- **Untracked row outside the caller's changed set**: overlay refuses (`baseline_scope_*` = 0), full-scan fallback, report == oracle, untracked contribution counted exactly once.
- **Unstaged deletion + staged edit elsewhere**: deleted file absent from rows and from the added-lines changed set, so overlay succeeds with the deleted module's contributions gone from aggregates; == oracle. Deletion handling is symmetric between overlay and legacy scan (both iterate present-on-disk files).
- **Staged pure rename**: conservative fallback (rename marks both paths dirty; old path has no added lines so sits outside the changed set); == oracle; counter math fully reconciled (4 fact hits, 3 fresh parses).
- **Clean branch switch**: after switching branches with unchanged tree content, the untouched sibling scope was served by identity alone (hits=1, computed=2) — reuse is fingerprint-driven, not path/state-driven.
- **Corruption matrix** (controlled per-mode): truncated → identical to oracle; garbled/non-JSON → identical (entry removed, recomputed); schema-old → identical (ignored, recomputed). Matches the unit tests `scope_cache_roundtrip_rejects_corrupt_and_foreign_entries` / `corrupt_baseline_cache_recomputes_without_altering_findings` / `version_old_baseline_cache_is_ignored_and_recomputed`.

## Audit conclusions

- **Exactness of dirty/staged/untracked/deleted/renamed rows**: `git_rust_inputs` marks *every* porcelain status record dirty (staged, unstaged, untracked, rename source+target) → `content_id: None`; a diff-HEAD-changed tracked file necessarily carries a status entry, so a row can never be simultaneously counted in the clean baseline and re-added as a changed occurrence — **no double-count is reachable**. Volatile rows outside the changed set and changed paths missing from rows both force deterministic full-scan fallback (src/test_module_aggregate.rs guard loops).
- **Subtraction/replacement**: `TestModulePrecedent::excluding(name)` removes exactly one occurrence of the target name; histogram-equivalent to the legacy identity-based exclusion (duplicate `(path,line,name)` triples are unreachable from a single parse). `total` stays consistent (`debug_assert` holds by induction). Changed-file occurrences enter the precedent exactly once, replacing (not adding to) their old-content contributions because volatile rows never join the scope tree.
- **Cache identity**: SHA-256 over framed material binding scheme tag, `CACHE_NAMESPACE`, `CACHE_SCHEMA_VERSION`, `AGGREGATE_SCHEMA_VERSION`, canonicalized absolute repository root, per-node relative coordinate, `SKIPPED_DIRS`, ordered file-name→blob-id pairs, and child Merkle fingerprints (BTreeMap ordering deterministic). Cross-checkout leakage blocked by root coordinate (unit-tested); aliasing impossible since all coordinates derive from git output with framed lengths.
- **Fallback robustness**: missing/truncated/garbled entries are deleted and recomputed; version-old and foreign-fingerprint entries ignored; cache-layer IO failures degrade to full scan; `Err(_)` from aggregation falls back deterministically. No panics observed in any corruption mode. Findings never altered except in the adversarial case below.
- **Privacy**: persisted scope entries contain only `schema_version`, hex fingerprint, aggregate module-name counts, total. Verified by inspecting real cache files: no paths, no source text, no line numbers, no finding text. (Module names are the same data class the pre-existing per-file fact cache already persists.)
- **Perf honesty**: counters reconcile from first principles in every scenario I ran; the warm path performs git metadata enumeration (`rev-parse`, `ls-files`, `ls-files -s`, `status`) but reads no file contents and touches no fact rows when the root scope hits. The PR's wording ("avoided repository-wide fact-row aggregation") is accurate; issue #50's stronger "cost scales with changed scopes rather than total repository file count" benchmark criterion remains honestly out of this slice's claim.
- **Atomicity/concurrency**: temp file + `fs::rename` per entry; cross-process races benign (deterministic identical payloads, last-rename-wins); `exists()` precheck avoids redundant writes; corrupt outcomes self-heal on next load. Same env controls as `FactCache`.

## Findings (ranked)

1. **LOW — well-formed wrong-data envelopes are trusted.** A structurally valid `ScopeEnvelope` with matching `schema_version` + `scope_fingerprint` but fabricated counts is accepted and altered findings in a controlled poison test (injected `evil_tests=999` changed the reported total). Not a regression: the pre-existing per-file `FactCache` trusts well-formed blobs keyed by content-id identically (verified same validation shape at src/rust_facts.rs `load`), the cache is user-local, and repository content cannot forge a fingerprint (binds canonical root + blob identities under SHA-256). Worth one sentence in the module doc's exactness contract and/or a future HMAC/content-recovery design note; not blocking for this slice.
2. **LOW — schema-old entries are never garbage-collected.** `ScopeCache::load` deletes only parse-failing files; version-old entries linger forever in `test-module-scopes-v1/`. Hygiene only; no correctness impact.
3. **INFO — counter lossiness on failure paths.** `record_baseline_scopes` fires only on overlay success; an aggregation that errors mid-way (e.g., a clean row vanishing between status and parse) reports zero scope activity despite work performed, then falls back. Honest but slightly lossy observability; the silent `Err(_) => {}` swallow is deliberate ("advisory") and safe.
4. **INFO — platform coverage.** Store path uses POSIX-friendly rename semantics; Windows rename-over-existing would skip storing (benign recompute), but CI exercises ubuntu only. No action required for this slice.
5. **Pre-existing boundaries (unchanged by this PR, for the record)**: TOCTOU between status snapshot and file read could cache disk bytes under a stale blob id, and skip-worktree/assume-unchanged files are treated as clean with index-blob identity — both behaviors exist identically at base through `scan_rust_repository`.

## Verdict

Exact head is current-base-descendant with zero main movement; all material receipt claims (patch-id, 181 focused tests, clippy/fmt, full-suite base-equivalent failure, cold/warm/fallback/corruption dogfood shapes and warm perf numbers) were independently reproduced; no correctness, privacy, or performance-boundary blocker stands. The low findings above are documentation/hygiene follow-ups, none affecting this slice's guarantees.

**ACCEPT**

— 🔬 Caliper · Ox portfolio pod

teamleaderleo:
Reviewed as 🔬 Caliper — implementation-independent reviewer, Ox portfolio pod. Fresh exact-head review of the repair; the prior ACCEPT of `259d07d` is explicitly not carried forward. Nothing below is taken from the implementation receipt without local reproduction.

## Reviewed coordinates

- Base: `b1a0c1301bcb556b1134782508070d6c3a40e088` == live `origin/main` (fresh fetch, zero movement)
- Head: `fcde3366c1e36892125f6898c4b2b2c5d5e26788` (detached clean worktree); PR head, base, and draft state verified against GitHub before work
- Ancestry: head descends from base via exactly one repair commit on the reviewed `259d07d`
- Hosted checks inspected only (none dispatched/rerun): `rust` pass ×2, `derived-root` pass, GitGuardian pass

## Seal preimage — independently re-derived

I re-implemented the seal in an independent Python implementation and compared it against real cache bytes written by the release binary at this exact head:

- Layout: framed domain tag `cultist-test-module-scope-seal-v2` → u32-LE schema version → framed scope fingerprint → u64-LE entry count → per entry (framed name, u64-LE count) in `BTreeMap` (key-sorted) order → u64-LE total.
- All three stored entries' seals matched my recomputation byte-for-byte: the encoding is deterministic, canonical (sorted-map iteration; JSON object order cannot influence the digest), injective/length-framed (sequential decoding is unambiguous, so no two distinct payloads share a preimage), domain-separated from the fingerprint scheme, and covers schema version, exact scope fingerprint (which transitively binds root coordinate, namespace, extractor/schema generation, selection config, child Merkle material), the full ordered `name_counts`, and `total`. All multi-byte encodings are explicit little-endian; no platform-endianness assumption exists.

## Attack matrix — all executed against the running binary with isolated caches

Every probe: findings byte-identical (`cmp`) to a no-cache oracle run, forged data never served.

| Probe | Result |
| --- | --- |
| counts-only tamper under stale seal | miss → quarantined → repaired |
| total-only tamper under stale seal | miss → quarantined → repaired |
| coherent counts+total fabrication under stale seal | miss → quarantined → repaired (also covered end-to-end by the new diff test) |
| honestly sealed but sum ≠ total | rejected by the checked-sum invariant alone → quarantined |
| u64-boundary overflow sum (2⁶³+2⁶³), honest seal | `checked_add` rejects, no panic/wrap |
| missing seal field / garbled seal / truncated bytes | deserialization or digest failure → quarantined |
| v1-shaped unsealed entry planted in the v2 namespace | fails deserialization (required seal field) → quarantined |
| duplicate `name_counts` JSON keys, forged copy last | serde last-wins value must re-seal → mismatch → quarantined |
| unknown extra fields, intact payload | inert: only seal-verified `name_counts`/`total` are returned; findings unchanged |
| float / negative numeric values | deserialization failure → quarantined |
| self-consistently sealed old-schema entry | ignored, kept (historical stale-entry semantics preserved) |
| self-consistently sealed foreign-coordinate entry | ignored, kept |

Quarantine-vs-ignore was discriminated behaviorally (quarantine ⇒ miss then hit after re-store; ignore ⇒ miss then miss) and matches the literal deletion assertions in the unit suite. Old-entry GC remains advisory exactly as permitted: ignored entries can permanently cost a recompute of that scope but can never affect results because both gates precede any trust.

## Independent dogfood replay (disposable scratch repo + isolated cache dir)

- Cold (staged edit): 7 parses, 0 fact hits, 3 scopes computed, 0 hits — report byte-identical to `CARGO_CULTIST_CACHE=0` oracle.
- Warm same-state: 1 parse (changed file only), root-scope identity hit 1/0 — byte-identical to oracle.
- Coherent fabrication of **all three** entries (+1000 `fabricated_tests` in counts and total, original seals kept): **0 hits, 3 recomputed**, report byte-identical to oracle.
- Repair run: 1 hit, 0 computed, byte-identical; grep found zero forged residue; I re-verified every post-repair file's seal and invariant with the independent implementation.
- Cache tree contained only the pre-existing fact namespace plus exactly three sealed `test-module-scopes-v2` entries, zero `.tmp` residue, and no unrelated persistent state. Real entry contents confirmed privacy: names/counts/total/fingerprint/seal only.

## Commands and results (exact head)

| Command | Result |
| --- | --- |
| `cargo fmt --check` | clean |
| `cargo clippy --all-targets -- -D warnings` | green |
| `cargo test --bins` | 187 passed, 0 failed (includes all six new adversarial tests) |
| `cargo test` (full) | 77/78 binaries ok; sole failure `rust_edit_class_source_current_main::focused_source_values_and_focus_admission_survive_current_main` (`pathspec 'master'`) reproduced identically by me in an isolated clone of exact base `b1a0c13` — pre-existing environment assumption, not introduced here |
| `git diff --check b1a0c13..HEAD` | clean |

## Boundary truthfulness

The module doc and PR body state the trust boundary precisely: an unkeyed, domain-separated content digest that detects accidental corruption and stale/partial writes, explicitly *not* authentication, with the malicious same-user writer who recomputes the digest declared out of scope and no HMAC claim made. The implementation does not exceed these claims anywhere I searched. The pre-existing per-file fact cache keeps its own unchanged trust model, as stated.

## Findings (ranked, none blocking)

1. **INFO — recovery-identity wording.** The PR's "stable patch-id `80133777…`" reproduces exactly for base→`259d07d` (I re-derived it), but the full base→`fcde336` diff now hashes to `d8ebdb43…` because of the repair commit. The sentence scopes itself to "the transplanted implementation," so it is not false, yet a future worker treating it as the whole-PR diff identity will mismatch. A follow-up receipt line stating the repaired-diff patch-id would remove the ambiguity.
2. **INFO — carried over, allowed by the lease.** Ignored old-schema/foreign-coordinate entries in the v2 namespace are never garbage-collected, so a littered entry permanently forces that scope's recompute (perf-only, correctness unaffected).
3. **INFO — quarantine is load-triggered.** A poisoned child entry beneath an intact root stays on disk until the root scope's identity changes; while the root hits, the child is never read, so it cannot alter results (observed directly during probing). Worth knowing when reasoning about residue.

## Verdict

Exact inputs verified unmoved; the v2 seal preimage was independently reproduced end-to-end; every invalid-cache shape I could construct — including coherent fabrications under stale seals and honestly-sealed invariant violations with overflow-boundary arithmetic — missed without altering findings, quarantined, and repaired deterministically; warm performance behavior and counters reconcile; the documented non-adversarial trust boundary is stated and implemented honestly; fmt/clippy/full-suite/diff hygiene all pass with the sole failure proven pre-existing on the exact base.

**ACCEPT**

— 🔬 Caliper · Ox portfolio pod

